Issue #115, opened 2014-12-23
Tornado 4 compatibility


ijl:
Tornado 4 isn't usable at the moment.

See https://travis-ci.org/bitly/pynsq/jobs/44946872.

```
[nsqd] 2014/12/23 15:47:33.754914 NSQ: persisting topic/channel metadata to nsqd.884.dat
ERROR:nsq.async:uncaught exception in data event
Traceback (most recent call last):
  File "/home/travis/build/bitly/pynsq/nsq/async.py", line 219, in _read_body
    self.trigger(event.DATA, conn=self, data=data)
  File "/home/travis/build/bitly/pynsq/nsq/event.py", line 85, in trigger
    ev(*args, **kwargs)
  File "/home/travis/build/bitly/pynsq/nsq/async.py", line 423, in _on_data
    self.trigger(event.RESPONSE, conn=self, data=data)
  File "/home/travis/build/bitly/pynsq/nsq/event.py", line 85, in trigger
    ev(*args, **kwargs)
  File "/home/travis/build/bitly/pynsq/nsq/async.py", line 385, in _on_response_continue
    self.trigger(event.READY, conn=self)
  File "/home/travis/build/bitly/pynsq/nsq/event.py", line 85, in trigger
    ev(*args, **kwargs)
  File "/home/travis/build/bitly/pynsq/tests/test_reader.py", line 151, in _on_ready
    c.send(nsq.subscribe(topic, 'ch'))
  File "/home/travis/build/bitly/pynsq/nsq/async.py", line 225, in send
    self.stream.write(data)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/tornado/iostream.py", line 348, in write
    self._handle_write()
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/tornado/iostream.py", line 1242, in _handle_write
    super(SSLIOStream, self)._handle_write()
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/tornado/iostream.py", line 802, in _handle_write
    self._write_buffer_size -= num_bytes
TypeError: unsupported operand type(s) for -=: 'int' and 'NoneType'
[nsqd] 2014/12/23 15:47:38.689345 PROTOCOL(V2): [127.0.0.1:50281] exiting ioloop
[nsqd] 2014/12/23 15:47:38.689432 ERROR: client(127.0.0.1:50281) - failed to read command - EOF
```

mpe:
[...]

PR #118, opened 2015-02-04
Pin to tornado<4 until compatible

ijl:
Let's pin installs to not exceed the last compatible version of Tornado until #115 is resolved.


mreiferson:
good idea

PR #121, opened 2015-03-10
Triggering the correct "CLOSE" event when the socket is closed

ellulpatrick:
I suspect this was a typo. The "CONNECT" event was being triggered when the socket was closed. It should instead be the "CLOSE" event.
Regards.


mreiferson:
thanks!

Issue #122, opened 2015-03-20
Reader.query_periodic may be None in Reader.close()


mpe:
If there is no lookupd configured then ```Reader.query_periodic``` will be ```None```, leading to:

```
File "/usr/local/lib/python2.7/dist-packages/nsq/reader.py", line 265, in close
     self.query_periodic.stop()
AttributeError: 'NoneType' object has no attribute 'stop'
```

mreiferson:
Hi @mpe - thanks for the report.  Any interest in submitting a fix?

mpe:
I'll try and get you one this week.

mreiferson:
fixed in #123

PR #123, opened 2015-03-25
Fix exception in Reader.close() when query_periodic is None

mpe:
Fixes bug #122.

Add a call to close to the message test to exercise the code. Verified
the test fails without the fix.


mreiferson:
thanks!

PR #124, opened 2015-03-27
Reenable tornado 4

mpe:
The Snappy & Deflate socket send routines need to return the number of
bytes they sent.

If the underlying socket.send() fails then we're toast.

Update travis to test with tornado 4.0 & 4.1, and update setup.py.


jehiah:
looks great. thanks!

PR #125, opened 2015-04-09
asyncconn: catch connection closed error on read_bytes

scottbelden:
If read_bytes is called when the connection is closed and there is nothing in the buffer, an iostream.StreamClosedError is thrown. This is caught in _read_size by the generic exception catcher but uncaught in _start_read.

This can be checked for beforehand by doing checking "if self.stream.closed()" before calling self.stream.read_bytes(). In this case, no buffered data would get processed, but processing the buffered data seems to be the intended behavior of the stream. Of course, processing the buffered data will eventually result in the error log message about not being able to send the FIN.

### DIFF
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -228,8 +228,19 @@ class AsyncConn(event.EventedMixin):
         self._start_read()
         self.trigger(event.CONNECT, conn=self)
 
+    def _read_bytes(self, size, callback):
+        try:
+            self.stream.read_bytes(size, callback)
+        except IOError:
+            self.close()
+            self.trigger(
+                event.ERROR,
+                conn=self,
+                error=protocol.ConnectionClosedError('Stream is closed'),
+            )
+
     def _start_read(self):
-        self.stream.read_bytes(4, self._read_size)
+        self._read_bytes(4, self._read_size)
 
     def _socket_close(self):
         self.state = DISCONNECTED
@@ -241,7 +252,6 @@ class AsyncConn(event.EventedMixin):
     def _read_size(self, data):
         try:
             size = struct.unpack('>l', data)[0]
-            self.stream.read_bytes(size, self._read_body)
         except Exception:
             self.close()
             self.trigger(
@@ -249,6 +259,8 @@ class AsyncConn(event.EventedMixin):
                 conn=self,
                 error=protocol.IntegrityError('failed to unpack size'),
             )
+            return
+        self._read_bytes(size, self._read_body)
 
     def _read_body(self, data):
         try:


mreiferson:
Good catch.

Seems like a cleaner way to resolve this would be to wrap `self.stream.read_bytes(...)` with a method on `AsyncConn` so we can catch these (and potentially other exceptions) in one spot?

scottbelden:
Ah, I guess I should have run the tests beforehand. I'll add the wrapper method and fix up the tests.

mreiferson:
Thanks @scottbelden, appreciate the contribution!

FYI, in the future, we're OK with pushing new commits to this branch/PR for review and then squashing down to logical commits before merging.

scottbelden:
@mreiferson While testing I noticed that the travis tests use tornado 2.4.1 (which didn't have StreamClosedError and therefore causes the failures).

https://github.com/tornadoweb/tornado/commit/3258726fea5bcd1b401907653bc953ce63d5aeb2

Would it be best to catch the IOStream error instead so that this is compatible with tornado 2.4.1?

mreiferson:
@scottbelden ahhh - we can either catch an error that exists across versions, if that's possible, or we can pivot on the exception to catch based on version (i.e. assign it to some global at import time)?

nsq/async.py
```diff
@@ -241,7 +251,7 @@ def close(self):
     def _read_size(self, data):
         try:
             size = struct.unpack('>l', data)[0]
-            self.stream.read_bytes(size, self._read_body)
+            self._read_bytes(size, self._read_body)
```
mreiferson: sorry, forgot about this PR :frowning:...

Since `_read_bytes` is now capturing the exception, doesn't that mean we're now going to miss calling the `self.close()` in the exception handler for this block?

scottbelden: Good point. Should _start_read() also call self.close() on exceptions? If so, we can just add self.close() to the _read_bytes exception handler.

mreiferson: yes, I think moving the close into `_read_bytes` is fine

nsq/async.py
```diff
@@ -241,7 +252,7 @@ def close(self):
...
         except Exception:
             self.close()
```
mreiferson: probably don't need this anymore

scottbelden: Based on the original code it looked like the exception was to handle struct unpacking and that self.close() was there if the struct unpacking threw an exception. So now if the unpack fails the stream should not be closed? I don't know enough about the low level details, but it would be a different behavior than the original code.

mreiferson: You're right, good call.

I still think we should clarify that though, perhaps by moving the `_read_bytes` call out of the exception scope?

scottbelden: Agreed. Moved that.

mreiferson:
@scottbelden tests are failing (need to return in that exception handler now), see:

https://travis-ci.org/bitly/pynsq/jobs/62449990

scottbelden:
@mreiferson I think this should be good to go. Let me know if you see anything else.

mreiferson:
@scottbelden thanks, LGTM - do you mind squashing down to one commit and we'll merge this in?

scottbelden:
@mreiferson Sure. I've never had to do that before? Is rebasing the way to go?

mreiferson:
@scottbelden yes, an "interactive" rebase will give you the option to select commits to squash - thanks!

scottbelden:
@mreiferson I've been busy lately but I think I did the rebase correctly. It's just re-testing now.

mreiferson:
ya, LGTM

thanks again @scottbelden

PR #117, opened 2015-05-12
nsq.Reader.disabled() no longer effective

jehiah:
Due to the "don't decrement rdy" change to nsqd behavior (https://github.com/bitly/nsq/pull/404), the current effect of `disabled()` returning `True` in `nsq.Reader`, of not sending any ready count (never calling `_send_rdy()`), fails to reduce the ready count to zero on the nsqd side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,21 @@
 language: python
 python:
   - 2.6.9
-  - 2.7.9
+  - 2.7
 env:
-  - NSQ_DOWNLOAD=nsq-0.2.28.linux-amd64.go1.2.1 TORNADO_VERSION=2.4.1
-  - NSQ_DOWNLOAD=nsq-0.2.28.linux-amd64.go1.2.1 TORNADO_VERSION=3.2
-  - NSQ_DOWNLOAD=nsq-0.2.28.linux-amd64.go1.2.1 TORNADO_VERSION=4.0
-  - NSQ_DOWNLOAD=nsq-0.2.28.linux-amd64.go1.2.1 TORNADO_VERSION=4.1
-  - NSQ_DOWNLOAD=nsq-0.2.30.linux-amd64.go1.3 TORNADO_VERSION=2.4.1
   - NSQ_DOWNLOAD=nsq-0.2.30.linux-amd64.go1.3 TORNADO_VERSION=3.2
   - NSQ_DOWNLOAD=nsq-0.2.30.linux-amd64.go1.3 TORNADO_VERSION=4.0
   - NSQ_DOWNLOAD=nsq-0.2.30.linux-amd64.go1.3 TORNADO_VERSION=4.1
-  - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=2.4.1
   - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=3.2
   - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=4.0
   - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=4.1
-  - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=2.4.1
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=3.2
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=4.0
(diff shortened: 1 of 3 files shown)


jehiah:
finally taking a pass at this

jehiah:
RFR @mreiferson (i'll confirm once i've validated this change set). 

I successfully moved the travis builds for this over to container based infra =)

test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
mreiferson: Typically `test.sh` doesn't have side effects like installing pkgs and such.  I'm fine if you want to organize code into a script for travis, but let's not call it `test.sh`?

.travis.yml
```diff
@@ -1,34 +1,21 @@
...
   - NSQ_DOWNLOAD=nsq-0.2.30.linux-amd64.go1.3 TORNADO_VERSION=4.0
   - NSQ_DOWNLOAD=nsq-0.2.30.linux-amd64.go1.3 TORNADO_VERSION=4.1
-  - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=2.4.1
   - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=3.2
   - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=4.0
   - NSQ_DOWNLOAD=nsq-0.2.31.linux-amd64.go1.3.1 TORNADO_VERSION=4.1
-  - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=2.4.1
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=3.2
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=4.0
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=4.1
```
mreiferson: are we missing a stable release between `0.3.0` and `0.3.5`? (looks like `0.3.1` and `0.3.2`)

mreiferson:
These changes look fine but it doesn't like it completely addresses the issue in the OP - `disabled()` is still ineffective.

I'm assuming that in your actual reader you're subclassing `disabled()` and explicitly calling `change_max_in_flight(0)`.  Isn't that something that `Reader` should do for you so that we preserve the semantics of `disabled()` returning a `bool` to indicate state?

test.sh
jehiah: fair point. updating to `travis_test.sh`

jehiah:
How i'm attempting to use this is [here](https://gist.github.com/jehiah/44b717b8a09e984ac470). which should expose enough to make disabling possible. I'm not 100% sure it'll work yet, but i think it will =)

I propose we add a check in disabled() such that it logs a warning if used against a newer nsqd where it's inefective. We can then remove it completely at the appropriate point when we drop support for older nsqd versions. Sound good?

mreiferson:
>I propose we add a check in disabled() such that it logs a warning if used against a newer nsqd where it's inefective. We can then remove it completely at the appropriate point when we drop support for older nsqd versions. Sound good?

OK, sounds reasonable.

jehiah:
@mreiferson previous comments addressed. Ready for :hammer: ?

.travis.yml
mreiferson: this one?

nsq/reader.py
```diff
@@ -692,13 +710,36 @@ def giving_up(self, message):
...
+        
+    @classmethod
+    def disabled(cls):
         """
         Called as part of RDY handling to identify whether this Reader has been disabled
 
         This is useful to subclass and override to examine a file on disk or a key in cache
         to identify if this reader should pause execution (during a deploy, etc.).
+        
+        Note: deprecated. Use change_max_in_flight(0)
```
mreiferson: we ended up calling the method `set_max_in_flight`

nsq/reader.py
```diff
@@ -692,13 +710,36 @@ def giving_up(self, message):
...
+            def semver(v):
+                def cast(x):
+                    try:
+                        return int(x)
+                    except:
+                        return x
+                return map(cast, v.replace('-','.').split('.'))
+
+            if self.disabled.__code__ != Reader.disabled.__code__ and semver(data['version']) >= semver('0.3'):
+                logging.warning('disabled() incompatible with nsqd >= 0.3. Use change_max_in_flight(0)')
```
mreiferson: same here (naming)

nsq/reader.py
```diff
@@ -692,13 +710,36 @@ def giving_up(self, message):
...
+                warnings.warn('disabled() is a deprecated and will be removed in a future. ' +
```
mreiferson: > disabled() is deprecated and will be removed in a future release, use set_max_in_flight(0) instead

jehiah: engrish

jehiah:
comments fixed and :hammer:'d

mreiferson:
LGTM

PR #127, opened 2015-06-12
Python 3

### LINKED ISSUE #107: Python 3
Is Python 3 support a goal? Would a patch be accepted?

ijl:
pynsq is compatible with Python 3 on this branch.

There are byte/unicode issues and import paths. 

I've cherry-pick @stanhu's patch for deflate and snappy sockets in order to use Tornado with Python 3.

Fixes #107.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - 2.6.9
   - 2.7.9
+  - 3.3
 env:
   - NSQ_DOWNLOAD=nsq-0.2.28.linux-amd64.go1.2.1 TORNADO_VERSION=2.4.1
   - NSQ_DOWNLOAD=nsq-0.2.28.linux-amd64.go1.2.1 TORNADO_VERSION=3.2
--- a/nsq/protocol.py
+++ b/nsq/protocol.py
@@ -9,10 +9,11 @@ except ImportError:
     import json  # pyflakes.ignore
 
 from .message import Message
+from nsq import compat
 
 
-MAGIC_V2 = '  V2'
-NL = '\n'
+MAGIC_V2 = compat.b('  V2')
+NL = compat.b('\n')
 
 
 FRAME_TYPE_RESPONSE = 0
@@ -21,16 +22,16 @@ FRAME_TYPE_MESSAGE = 2
 
 
 # commmands
-AUTH = 'AUTH'
-FIN = 'FIN'  # success
-IDENTIFY = 'IDENTIFY'
-MPUB = 'MPUB'
-NOP = 'NOP'
-PUB = 'PUB'  # publish
-RDY = 'RDY'
-REQ = 'REQ'  # requeue
-SUB = 'SUB'
-TOUCH = 'TOUCH'
+AUTH = compat.b('AUTH')
+FIN = compat.b('FIN')  # success
+IDENTIFY = compat.b('IDENTIFY')
+MPUB = compat.b('MPUB')
+NOP = compat.b('NOP')
+PUB = compat.b('PUB')  # publish
+RDY = compat.b('RDY')
+REQ = compat.b('REQ')  # requeue
+SUB = compat.b('SUB')
+TOUCH = compat.b('TOUCH')
 
 
 class Error(Exception):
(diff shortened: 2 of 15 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 python:
   - 2.6.9
   - 2.7.9
+  - 3.4
```
ijl: 3.4.3 wasn't available... I can pin to 3.4.2 if this build's green and you prefer.

mreiferson: this is fine

tests/test_reader.py
```diff
@@ -206,7 +206,7 @@ class DeflateReaderIntegrationTest(ReaderIntegrationTest):
         'deflate_level': 6,
         'tls_v1': True,
         'tls_options': {'cert_reqs': ssl.CERT_NONE},
-        'heartbeat_interval': 10,
+        'heartbeat_interval': 1,
```
mreiferson: this change looked odd, any particular reason?

ijl: If I remember correctly, it was for a faster test while debugging. Like it reverted?

nsq/protocol.py
```diff
@@ -70,15 +71,24 @@ def decode_message(data):
...
-    body_data = ''
-    params_data = ''
+    body_prefix = b''
+    params_data = b''
     if body:
-        assert isinstance(body, str), 'body must be a string'
-        body_data = struct.pack('>l', len(body)) + body
+        assert isinstance(body, compat.string_like), 'body must be a string'
+        if isinstance(body, compat.unicode):
+            body = body.encode('utf-8')
```
mreiferson: We had a discussion about an approach like this a long time ago, I'm still of the opinion that we should not be encoding the body passed in here and instead should assert bytes (or `str`).

nsq/writer.py
```diff
@@ -109,10 +108,35 @@ def _run(self):
...
+        :param str topic: The topic to publish to.
+
+        :param msg: Message payload. If this is not bytes or an iterable
+            of bytes, it is assumed to be a UTF-8 encoded string or an
+            iterable of UTF-8 encoded strings.
+
+        :type msg: str, unicode, or bytes
+
+        :param callable callback: Callback invoked on response or error.
+        """
```
mreiferson: thanks for these :+1:

tests/test_reader.py
```diff
@@ -28,7 +28,7 @@ class ReaderIntegrationTest(tornado.testing.AsyncTestCase):
         'snappy': True,
         'tls_v1': True,
         'tls_options': {'cert_reqs': ssl.CERT_NONE},
-        'heartbeat_interval': 10,
+        'heartbeat_interval': 1,
```
mreiferson: same here?

mreiferson:
waiting for travis, but this LGTM other than the minor comments above, thanks for taking care of this @ijl :heart_eyes:

tests/test_reader.py
mreiferson: that's a good reason, let's keep it!

nsq/protocol.py
ijl: Yeah, I saw. I instead documented the casting in `Writer.pub` and `Writer.mpub` (there may be other places, as well?), which was mentioned in the previous discussion. 

But I can require bytes if you like, though I would tend to think that this will handle the most common case conveniently. If you have unicode or bytes, as you almost certainly do, it works fine; if you don't, read the docs and pass in bytes.

mreiferson: Now that it's documented I think I can live with it

ijl:
@mreiferson apparently Python 3.4.2 has a bug in mock (fixed in 3.4.3), so I've tested on 3.3. You'll see nsq 0.2.28 fails. Per #117, are you dropping support for that?

mreiferson:
Not exactly _officially_ dropping support but the reason the test is failing is odd, though:

```
E_BAD_TOPIC PUB topic name 'test_conn_suscribe_1434151065.1826334' is not valid
```

Other test combinations against `0.2.28` are passing, which leads me to believe it doesn't have anything to do with that version of `nsqd` and more to do with something on this side?

mreiferson:
let's keep this going, see #150

PR #130, opened 2015-08-30
Add SIGINT handler

jonmorehouse:
This adds in a SIGINT handler, to ensure that ctrl-c, is caught in the
nsq.run() loop.

@mreiferson this is a follow up from our conversation, let me know if there's anything else you'd like to see documented and/or anything else that needs to be done to get this pushed through.


mreiferson:
:+1: 

@jehiah do you recall any reason why we didn't add this initially? (I couldn't remember)

ploxiln:
One possible reason I can think of is so there's an easy way to "terminate faster", but still let python clean itself up and maybe print a backtrace. That might not be a great reason though. Maybe it was just an oversight.

PR #132, opened 2015-11-12
reader: message timeout callbacks

akellehe:


.gitignore
```diff
@@ -3,3 +3,4 @@ dist
 pynsq.egg-info
 build
 docs/_build
+.idea
```
akellehe: Probably want to take this out... it's for PyCharm

nsq/async.py
```diff
@@ -468,6 +468,9 @@ def _on_data(self, data, **kwargs):
             self.in_flight += 1
 
             message = protocol.decode_message(data)
+            if self.msg_timeout:
+                message._set_timeout(self)
```
akellehe: Accessing the protected method here isn't ideal. Trying to convey to the user that `_set_timeout` shouldn't be called in their implementations.

nsq/event.py
```diff
@@ -28,6 +28,7 @@ class InvalidListenerError(Exception):
 TOUCH = 'touch'
 FINISH = 'finish'
 MESSAGE = 'message'
+MESSAGE_TIMEOUT = 'message_timeout'
```
akellehe: We could shorten this to `TIMEOUT` and `"timeout"` since it's the only timeout event... Would be less descriptive, though.

nsq/message.py
```diff
@@ -55,9 +56,43 @@ def __init__(self, id, body, timestamp, attempts):
...
+        """
+        self.conn = async_conn
+        self._clear_timeout()
+
+        def callback():
+            async_conn.trigger(event.MESSAGE_TIMEOUT, conn=self, message=self)
+            if self.timeout.is_running():
+                self._clear_timeout()
+
+        self.timeout = tornado.ioloop.PeriodicCallback(
```
akellehe: Using [`tornado.ioloop.PeriodicCallback`][1] instead of [`tornado.ioloop.IOLoop.add_timeout`][2] for the `ms` resolution. We can use `add_timeout` if we pass the delta as a float, though. 

[1]: http://tornadokevinlee.readthedocs.org/en/latest/ioloop.html#tornado.ioloop.PeriodicCallback
[2]: http://tornadokevinlee.readthedocs.org/en/latest/ioloop.html#tornado.ioloop.IOLoop.add_timeout

nsq/reader.py
```diff
@@ -112,6 +112,10 @@ def process_message(message):
     :param max_in_flight: the maximum number of messages this reader will pipeline for processing.
         this value will be divided evenly amongst the configured/discovered nsqd producers
 
+    :param message_timeout_handler: A callable that runs when a message times out. The parameters are
```
akellehe: Should we add some docs for `msg_timeout` here?

nsq/async.py
akellehe: As an alternative here we could create a registry for the timeouts here... then we wouldn't have to incorporate them into the state of the message. Another not-so-nice alternative would be to make the method public.

nsq/reader.py
```diff
@@ -275,6 +283,18 @@ def set_message_handler(self, message_handler):
         assert callable(message_handler), 'message_handler must be callable'
         self.message_handler = message_handler
 
+    def set_message_timeout_handler(self, message_timeout_handler):
+        """
+        Assigns the callback method to be executed for each message timeout that occurs. This method can only be used
+            in conjunction with the client-side msg_timeout setting. The client, otherwise, has no awareness of the
+            message's lifespan.
```
mreiferson: `nsqd` delivers the configured `msg_timeout` to the client in the response body of the `IDENTIFY` command, so the client always knows what the `msg_timeout` is.

akellehe: Awesome! That makes this feature _much_ more intuitive.

nsq/message.py
mreiferson: I feel like this should be implemented similar to the way the other message-based events are implemented, specifically that it triggers an event here that the parent connection is listening for, that can then call the appropriate callbacks.  I think that would reduce the state we're tracking here?

nsq/async.py
mreiferson: (and would probably obviate the need for this method at all)

akellehe:
@mreiferson what do you think about this approach? 
* Removed `_set_timeout` method from :class:`pynsq.message.Message`
* Added `timed_out` public attribute there.
* Added private `__timeouts` registry to :class:`nsq.async.AsyncConn`
* Changed to `add_timeout` interface and moved to :class:`nsq.async.AsyncConn`

nsq/message.py
```diff
@@ -51,13 +51,22 @@ class Message(event.EventedMixin):
...
+        self.timed_out = False
 
         super(Message, self).__init__()
 
+    def is_alive(self):
+        """
+        Indicates whether or not this :class:`nsq.message.Message` instance is timed out based on the
+            msg_timeout parameter.
+        """
+        return self.timed_out
```
akellehe: We may want to use `not self.timed_out and not self._has_responded` here.

nsq/message.py
```diff
@@ -51,13 +51,22 @@ class Message(event.EventedMixin):
...
+        self.timed_out = False
 
         super(Message, self).__init__()
 
+    def is_alive(self):
+        """
+        Indicates whether or not this :class:`nsq.message.Message` instance is timed out based on the
+            msg_timeout parameter.
+        """
+        return not self.timed_out
```
akellehe: We may want to use `not self.timed_out and not self._has_responded` here for a more comprehensive check

nsq/message.py
```diff
@@ -51,13 +51,22 @@ class Message(event.EventedMixin):
...
+
         self.id = id
         self.body = body
         self.timestamp = timestamp
         self.attempts = attempts
+        self.timed_out = False
 
         super(Message, self).__init__()
 
+    def is_alive(self):
```
mreiferson: I don't see this method used anywhere?  Do we need it?

nsq/message.py
```diff
@@ -51,13 +51,22 @@ class Message(event.EventedMixin):
     def __init__(self, id, body, timestamp, attempts):
         self._async_enabled = False
         self._has_responded = False
+
         self.id = id
         self.body = body
         self.timestamp = timestamp
         self.attempts = attempts
+        self.timed_out = False
```
mreiferson: If we don't need `is_alive`, then we don't need this variable either...

nsq/async.py
```diff
@@ -468,9 +471,18 @@ def _on_data(self, data, **kwargs):
             self.in_flight += 1
 
             message = protocol.decode_message(data)
+            if self.msg_timeout:
+                if message.id in self.__timeouts:
```
mreiferson: Shouldn't it be impossible to have a duplicate message in `self.__timeouts`?  In order to see the same message ID a message would need to have:

1. been explicitly requeued (in which case the timeout callback needs to be cleaned up)
2. timed out and been redelivered (in which case the timeout callback would have fired and been cleaned up)

So, I don't think we need this check?

nsq/reader.py
```diff
@@ -275,6 +282,15 @@ def set_message_handler(self, message_handler):
         assert callable(message_handler), 'message_handler must be callable'
         self.message_handler = message_handler
 
+    def set_message_timeout_handler(self, message_timeout_handler):
```
mreiferson: We only need `set_message_handler` because in certain cases you want to pass your `nsq.Reader` instance to your handler, so it needs to be instantiated before you can be passed in.

Is that why we need this `set_message_timeout_handler`?

nsq/async.py
```diff
@@ -468,9 +471,18 @@ def _on_data(self, data, **kwargs):
...
+                    self.io_loop.remove_timeout(self.__timeouts[message.id])
+
+                self.__timeouts[message.id] = self.io_loop.add_timeout(
+                    self.msg_timeout/1000., lambda: self.trigger(event.MESSAGE_TIMEOUT, self, message)
+                )
+
             message.on(event.FINISH, self._on_message_finish)
             message.on(event.REQUEUE, self._on_message_requeue)
             message.on(event.TOUCH, self._on_message_touch)
+            message.on(event.TIMEOUT, self._on_message_timeout)
```
mreiferson: `s/TIMEOUT/MESSAGE_TIMEOUT`

nsq/reader.py
```diff
@@ -476,6 +499,7 @@ def connect_to_nsqd(self, host, port):
         conn.on('close', self._on_connection_close)
         conn.on('ready', self._on_connection_ready)
         conn.on('message', self._on_message)
+        conn.on('message_timeout', self._on_message_timeout)
```
mreiferson: I don't see `AsyncConn` propagating this event, so I don't think this is listening to anything.  Maybe `_on_message_timeout` should be implemented at the `AsyncConn`, not `Reader`, level?

nsq/async.py
akellehe: Yeah, I think you're right

nsq/message.py
akellehe: This method offers a second approach to users for acting on the fact their message has timed out. If, for example, a message is timed out and about to undergo an expensive operation; the user can check `is_alive` and simply discard the message. 

The alternative using callbacks would require some significant overhead: either a registry of messages that have timed out so the same thing can be accomplished, or setting a similar flag in the callback itself. Imho neither of those two are ideal, so offering this interface gives the user a simple alternative.

nsq/message.py
akellehe: Worth noting: I think we could change the name from `is_alive` to `is_in_flight` if we include both flags.

nsq/reader.py
akellehe: Nope! I was just following the same pattern. I'll change this to a simple `=`

nsq/reader.py
akellehe: Wouldn't [this trigger][1] propagate to this callback?

[1]: https://github.com/akellehe/pynsq/blob/message-timeout-callback-and-state-accessor/nsq/async.py#L479

akellehe: Added [a test][1] to confirm

[1]: https://github.com/akellehe/pynsq/blob/message-timeout-callback-and-state-accessor/tests/test_reader.py#L201-L223

nsq/message.py
mreiferson: OK, but I don't think it needs to be implemented by relying on some parent to set a property on this object - the message could gain the knowledge of its timeout and then just perform some time logic here?

nsq/async.py
```diff
@@ -520,3 +530,20 @@ def _on_message_touch(self, message, **kwargs):
                 conn=self,
                 error=protocol.SendError('failed to send TOUCH %s' % message.id, e),
             )
+
+    def __clear_timeout(self, message):
```
mreiferson: I think all of these related timeout functions and variables need to be more specifically named, i.e. `s/timeout/message_timeout`

nsq/message.py
akellehe: That's a good point. I'll take a crack at something less leaky.

nsq/async.py
akellehe: +1

akellehe:
@mreiferson I think we're all set. Are these tests adequate? Should we test a few more cases?

mreiferson:
@akellehe Nice work so far!

I have a general concern that for high volume consumers there could be non-trivial overhead maintaining timers for every message received.

As implemented, this is going to track timeouts for all messages even if a message timeout handler _hasn't_ been provided.  Should we only track timeouts when a handler is provided?   If so, that means that `Message.is_in_flight()` needs to track state independent of message timeout events.

akellehe:
@mreiferson these are good points. LMK what you think about this approach to passing the responsibility for handling timeout state to the `nsq.message.Message`

nsq/protocol.py
```diff
@@ -61,12 +61,12 @@ def unpack_response(data):
     return frame, data[4:]
 
 
-def decode_message(data):
+def decode_message(data, timeout=None):
```
akellehe: I don't think setting the timeout here is ideal, `decode_message` isn't very idiomatic when we also use it to pass the message `timeout`. The best alternative I can think of is to keep track of timeouts by default, but we'd see the additional overhead you mentioned.

mreiferson: I think I would prefer a method `_set_timeout` instead?

PR #133, opened 2015-11-17
change logging level of RDY-redist and heartbeats to debug

ploxiln:
A very subjective change which I offer for comment.

Another idea is to allow passing a logging function to use for some of these, which defaults to `logging.debug` or `logging.info`, but which you could set to whatever.

### DIFF
--- a/nsq/client.py
+++ b/nsq/client.py
@@ -67,7 +67,7 @@ class Client(object):
             conn.close()
 
     def _on_heartbeat(self, conn):
-        logger.info('[%s:%s] received heartbeat' % (conn.id, self.name))
+        logger.debug('[%s:%s] received heartbeat', conn.id, self.name)
         self.heartbeat(conn)
 
     def heartbeat(self, conn):
--- a/nsq/reader.py
+++ b/nsq/reader.py
@@ -316,14 +316,14 @@ class Reader(Client):
         rdy_conn = conn
         if len(self.conns) > self.max_in_flight and time.time() - self.random_rdy_ts > 30:
             # if all connections aren't getting RDY
-            # occsionally randomize which connection gets RDY
+            # occasionally randomize which connection gets RDY
             self.random_rdy_ts = time.time()
             conns_with_no_rdy = [c for c in self.conns.itervalues() if not c.rdy]
             if conns_with_no_rdy:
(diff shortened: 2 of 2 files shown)


mreiferson:
I feel like the times where you would want to see these messages are _after_ you've experienced some issue you'd like to debug (and it would be too late to switch to debug).

Are they that voluminous?

ploxiln:
There are a few logging messages in pynsq at the debug level, which that logic would also apply to. If you were worried about that, you'd probably set logging to debug. I think that for these cases, if I'm suspicious about heartbeats or rdy distributing, I'll go check in nsqadmin (and then if that isn't illuminating and the problem is persistent, switch logging to debug).

That said ... this is just a band-aid for some annoyances of my current logging situation. All log lines from all services are sent to a hosted service which charges by the GiB, and I can't easily view a certain severity and above, particularly if some messages are multi-line, like backtraces. I need to find the time to bite the bullet and set up some sort of smarter logging cluster and integration for services in a few languages.

So, yeah, I guess these two lines are not really a significant problem.

mreiferson:
> I think that for these cases, if I'm suspicious about heartbeats or rdy distributing, I'll go check in nsqadmin (and then if that isn't illuminating and the problem is persistent, switch logging to debug).

I'm imagining the scenarios where clients have already disconnected and your only hope is the logs to try and piece together what happened.

But, you're right, I suppose the logic applies to the already existing debug log lines.  I _think_ those generate more noise and are a little less valuable, at least in my experience.

I'm open to ideas where this could be more configurable/controllable by the end-user.

ploxiln:
I think @jehiah expressed the sentiment at one point that "maybe libraries shouldn't do logging themselves" or something like that. Though this is a bit framework-y so it's hard to avoid. So I just now thought that the thing to do is be able to pass a logger to override the python default. But now that I think of that, I should be able to do something like

```python
logging.getLogger("nsq").setLevel(logging.WARNING)
```

so I'll go ahead and close this PR :)

ploxiln:
meh it's annoying that I can't, with the technique above, get the "connecting to nsqd" and "IDENTIFY sent/received" messages, but not the heartbeats and distributing RDYs. such is life I guess

PR #135, opened 2015-11-23
reader: by default, archive failed messages to disk

jehiah:
This is a follow up to #134 and changes the default behaviour of `Reader.give_up()` to be less surprising by writing messages that are "given up on" to disk instead of dropping them completely.

### DIFF
--- a/nsq/reader.py
+++ b/nsq/reader.py
@@ -9,6 +9,8 @@ import urlparse
 import cgi
 import warnings
 import inspect
+import datetime
+import os.path
 
 try:
     import simplejson as json
@@ -131,6 +133,8 @@ class Reader(Client):
 
     :param max_backoff_duration: the maximum time we will allow a backoff state to last in seconds
 
+    :param data_path: the directory failed messages will be archived to after ``max_tries``
+
     :param \*\*kwargs: passed to :class:`nsq.AsyncConn` initialization
     """
     def __init__(
@@ -149,6 +153,7 @@ class Reader(Client):
             lookupd_poll_jitter=0.3,
             lookupd_connect_timeout=1,
             lookupd_request_timeout=2,
+            data_path=None,
             **kwargs):
         super(Reader, self).__init__(**kwargs)
 
@@ -161,6 +166,7 @@ class Reader(Client):
         assert isinstance(lookupd_poll_jitter, float)
         assert isinstance(lookupd_connect_timeout, int)
         assert isinstance(lookupd_request_timeout, int)
+        assert isinstance(data_path, (str, unicode, None.__class__))
 
         assert lookupd_poll_jitter >= 0 and lookupd_poll_jitter <= 1
 
@@ -222,6 +228,7 @@ class Reader(Client):
 
         self.redist_periodic = None
         self.query_periodic = None
+        self.data_path = data_path
 
     def _run(self):
         assert self.message_handler, "you must specify the Reader's message_handler"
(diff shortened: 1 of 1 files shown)


jehiah:
RFR @mreiferson. I'd love any naming comments you have, and if you can think of an obvious way to disable this (an additional parameter of `archive_abandoned_messages=True` ?)

nsq/reader.py
```diff
@@ -702,14 +709,31 @@ def process_message(self, message):
...
 
-        This is useful to subclass and override to perform a task (such as writing to disk, etc.)
+        This is useful to subclass and override to perform a custom task (such as writing to disk, etc.)
 
         :param message: the :class:`nsq.Message` received
         """
-        logger.warning('[%s] giving up on message %s after %d tries (max:%d) %r',
-                       self.name, message.id, message.attempts, self.max_tries, message.body)
+
+        self.failed_count += 1
```
mreiferson: I don't see where this is being initialized?

nsq/reader.py
```diff
@@ -702,14 +709,31 @@ def process_message(self, message):
...
+        
+        date_str = datetime.datetime.utcnow().strftime("%Y%m%d-%H%M%S-%f")
+        filename = "%s_%d.failed.msg" % (date_str, self.failed_count)
+        filename = os.path.join(path, filename)
+        
+        logging.warning('[%s] giving up on message %s after %d tries (max:%d). Archived to %s %r',
+                        self.name, message.id, message.attempts, self.max_tries, filename, message.body)
+        
+        f = open(filename, 'wb')
+        f.write(message.body + '\n')
```
mreiferson: not sure we should append anything to the `message.body`, in the case of a binary message format

nsq/reader.py
```diff
@@ -702,14 +709,31 @@ def process_message(self, message):
...
-                       self.name, message.id, message.attempts, self.max_tries, message.body)
+
+        self.failed_count += 1
+        
+        path = os.path.join(self.data_path or "", self.topic + '_' + self.channel)
+        if not os.path.exists(path):
+            os.makedirs(path)
+        
+        date_str = datetime.datetime.utcnow().strftime("%Y%m%d-%H%M%S-%f")
+        filename = "%s_%d.failed.msg" % (date_str, self.failed_count)
```
mreiferson: I feel like `message.id` should be in the filename?

mreiferson: (too)

nsq/reader.py
```diff
@@ -702,14 +709,31 @@ def process_message(self, message):
...
+        This is useful to subclass and override to perform a custom task (such as writing to disk, etc.)
 
         :param message: the :class:`nsq.Message` received
         """
-        logger.warning('[%s] giving up on message %s after %d tries (max:%d) %r',
-                       self.name, message.id, message.attempts, self.max_tries, message.body)
+
+        self.failed_count += 1
+        
+        path = os.path.join(self.data_path or "", self.topic + '_' + self.channel)
```
mreiferson: We should probably use a separator that isn't a valid character for a topic/channel.  Perhaps the channel should just be a real subdir?

mreiferson:
@jehiah should we also bump `max_attempts` to `25`?

mreiferson:
Do we need to provide a param to disable archiving when we already encourage users to subclass and override the method?

nsq/reader.py
jehiah: haha. just copy & pasting from my custom `giving_up()` here. I'll try some real validation =)

nsq/reader.py
jehiah: yeah... this is a tough one. I'm certainly biased by my own `json` msg usage, but any newline expectations (like from nsq_to_file, nsq_tail, etc) are carried through here. Having this means you can `cat *.msg | to_nsq`, etc. 

I think there are three basic options

a) document and move on, 
b) remove newline
c) add a parameter to disable.

I'm inlined to do `a` or `c`

nsq/reader.py
jehiah: i have mixed feelings. In some ways it could augment `self.failed_count` (which has some downsides across multiple).

however since `message.id` is source unique not globally unique (depends on nsqd workerID) putting in the filename might imply uniqueness that it shouldn't. Also, even though we create hex message id's, the protocol defines it as arbitrary bytes which wouldn't be safe for filename.

I chose to calculate the filename and include that in the log line as an alternative way a) raise awareness of the file archiving and b) link the two.

mreiferson: > I chose to calculate the filename and include that in the log line as an alternative way a) raise awareness of the file archiving and b) link the two.

Fair point, that was the only reason why I thought it would be useful in the filename.

nsq/reader.py
mreiferson: Not a fan of more options (c) and I suppose that they can still override this method if one doesn't want a newline appended, so let's do (a)?

mreiferson:
what're we doing on this one?

Issue #139, opened 2015-12-20
support DPUB


iameugenejo:
This feature has been out since 0.3.6 (Sept 25, 2015) and I see this implemented in go client, but not in python one.

Any plan for supporting dpub? 

reference - https://github.com/nsqio/go-nsq/blob/37e5a55ed8546a1954c0dccf4e44a48e67db931b/command.go#L136-L141

mreiferson:
done in #140

iameugenejo:
awesome, thanks!

PR #140, opened 2015-12-20
#139 support DPUB

iameugenejo:


nsq/writer.py
```diff
@@ -57,6 +57,7 @@ def get(self):
 
                 self.nsq.pub(topic, msg) # pub
                 self.nsq.mpub(topic, [msg, msg_cn]) # mpub
+                self.nq.dpub(topic, 60, msg) # dpub
```
mreiferson: `nsq` typo

nsq/writer.py
```diff
@@ -131,7 +135,7 @@ def _pub(self, command, topic, msg, callback):
         conn.callback_queue.append(callback)
         cmd = getattr(protocol, command)
         try:
-            conn.send(cmd(topic, msg))
+            conn.send(cmd(topic, msg) if delay_ms is None else cmd(topic, delay_ms, msg))
```
mreiferson: this is unfortunate, we could probably clean it up, but it doesn't matter that much I guess

mreiferson:
@iameugenejo thanks for putting this together!

nsq/writer.py
iameugenejo: I tried my best not to modify much from original code. I can try cleaning it up.

mreiferson:
LGTM, would you mind squashing down to one commit?

Thanks!

iameugenejo:
The commit has been squashed.

Thank you!

Issue #143, opened 2016-02-05
Allow pause and resume Reader


iameugenejo:
I found a go client implementation it, but not in python client - https://github.com/nsqio/go-nsq/blob/master/consumer.go#L817-L855

I'll send a PR for it in a moment.

mreiferson:
:+1:

iameugenejo:
Please ignore the first two commits, I overwrote my own changes from another PR and didn't know until I pushed the new changes (did that twice), and deleted the branch thinking it would clear up the history. I guess github keep tracks of forked repositories' branches.

mreiferson:
https://github.com/nsqio/pynsq/pull/145#issuecomment-180464576

PR #145, opened 2016-02-05
#143 allow manual backoff and resume

iameugenejo:
- modified ```_start_backoff_block``` to take duration to override backoff timer interval when manually backing off
- added ```backoff()``` and ```resume()``` for manual backoff control
- added a unittest method for manual backoff control

### DIFF
--- a/tests/test_backoff.py
+++ b/tests/test_backoff.py
@@ -485,3 +485,104 @@ def test_backoff_conns_disconnect():
         for i, call in enumerate(c.stream.write.call_args_list):
             print "%d: %s" % (i, call)
         assert c.stream.write.call_args_list == [((arg,),) for arg in c.expected_args]
+
+
+def test_manual_backoff():
+    mock_ioloop = create_autospec(IOLoop)
+    r = _get_reader(mock_ioloop)
+    num_conns = 5
+    conns = []
+    for i in range(num_conns):
+        conn = _get_conn(r)
+        conn.expected_args = ['SUB test test\n', 'RDY 1\n']
+        conn.fails = 0
+        conns.append(conn)
+
+    conn = conns[0]
+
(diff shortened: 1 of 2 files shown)


tests/test_backoff.py
```diff
@@ -485,3 +485,104 @@ def test_backoff_conns_disconnect():
...
+    assert get_ready_counts(conns, call_arg_count_map) == 5, 'resume from manual backoff must send full ready count'
+
+    now = time.time()
+    backoff_duration = 99
+    r.backoff(backoff_duration)
+
+    assert r.backoff_block is True
+    assert mock_ioloop.add_timeout.called
+    timeout_args, timeout_kwargs = mock_ioloop.add_timeout.call_args
+    assert are_floats_close(timeout_args[0], now + backoff_duration), \
```
ploxiln: Now is seconds since unix epoch, which is a pretty big number. are_floats_close() might not be the best way to test timing tolerance.

are_floats_close() currently allows a "relative tolerance" of 1e-9 which comes out to about 1.45 seconds for time_t right now. That's not much, but consider how odd it is that the tolerance allowed will increase over the months and years. It would be simpler to compare to a specific time tolerance (and are_floats_close() could be removed)

mreiferson:
@iameugenejo thanks for the PR, but this is something that should be handled outside of `pynsq` by calling [`Reader.set_max_in_flight(...)`](https://github.com/nsqio/pynsq/blob/master/nsq/reader.py#L608-L622).

You pass a `0` to "pause" and pass non-zero (presumably the original value) to "resume", so it can be handled entirely outside of `pynsq`.

iameugenejo:
aha! I didn't see that.

I should've read through documentation first.

Thanks!

Issue #131, opened 2015-09-30
max_rdy_count hard-coded to 2500


bctsui:
In version 0.6.8:

AsyncConn.max_rdy_count is initialized to 2500 (line 175 in async.py) but the value from the actual nsqd server is never used. This means that Readers are effectively capped at 2500 max_in_flight.

It appears that the fix would be to parse max_rdy_count in AsyncConn._on_identify_response.

Reproduction steps:

1. `nsqd --max-rdy-count 10000 [...]`
2. Create a Python consumer: `nsq.Reader(..., max_in_flight=8000)`
3. Upon running the consumer, the following log message will appear: `WARNING:nsq.reader:[127.0.0.1:9999:POSTBACK] max RDY count 2500 < reader max in flight 8000, truncation possible`

mreiferson:
@bctsui thanks for the report!

I feel like this is a regression, but I can't find a trace of it in the commit logs.  Any interest in submitting a fix?

ploxiln:
twas in

commit 6c3e4f15a09c836a14fbd4a80267cd4aa94ab749
Author: Matt Reiferson <mreiferson@gmail.com>
Date:   Wed Nov 20 14:16:10 2013 -0500

    reader/writer: evented refactoring

mreiferson:
I know where that line was added, I was wondering if we ever had code to parse and update the value based on the response from `IDENTIFY`

ploxiln:
yup it's in there (that line was moved from further away, and the line of interest wasn't re-added)

https://github.com/nsqio/pynsq/commit/6c3e4f15a09c836a14fbd4a80267cd4aa94ab749#diff-2a8bc85bf9c95f482da1eb0490a60251L545

mreiferson:
doh!  nice find plo

mreiferson:
fixed in #146

PR #146, opened 2016-02-05
set max_rdy_count to value received from IDENTIFY response. see #131

klucar:
This fixes the hardcoding of max_rdy_count to 2500 to the value received from the IDENTIFY request. It defaults back to 2500 for older nsq versions. See #131


mreiferson:
LGTM, thanks!

PR #147, opened 2016-02-14
Work around a bug in mock's autospec support

nickstenning:
In order to get the tests passing with recent versions of mock, we need
to omit the `autospec=True` parameter for `socket.socket` mocks.

See:

  https://github.com/testing-cabal/mock/issues/323

### DIFF
--- a/tests/test_async.py
+++ b/tests/test_async.py
@@ -29,7 +29,12 @@ def _get_test_conn(io_loop=None):
     return conn
 
 
-@patch('nsq.async.socket', autospec=True)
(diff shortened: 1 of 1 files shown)


mreiferson:
should we pin the version of mock instead? what does this mean for _older_ versions of mock?

nickstenning:
Pinning the version of mock wouldn't be a bad thing, but this should be totally compatible with previous versions of mock. The only difference is that the `socket` mock will be slightly less strict (you'll be able to access attributes on it in the test that don't exist on the real thing.

mreiferson:
SGTM

PR #150, opened 2016-02-14
Python 3 compatibility, round two

nickstenning:
I'm super keen to get a version of `pynsq` working on Python 3. There appear to have been a couple of valiant attempts to get this working in the past, including #88, #102, and most recently #127. None of these have been merged, so here's another attempt to get this working, which tries to achieve the following:

- Each set of changes has been made in its own commit, with an explanation of what the change is and why it's needed for Python 3 compatibility.
- I've made sure that the concerns previously raised [[1](https://github.com/nsqio/pynsq/pull/102#discussion_r20363807), [2](https://github.com/nsqio/pynsq/pull/127#discussion_r32329639)] around ensuring that message bodies are bytestrings, not text, have been addressed. (*N.B.* The code previously didn't check individual bodies in `mpub` until they were sent as one fully-encoded command. I've [added an assertion](https://github.com/nsqio/pynsq/pull/150/files#diff-7be937e744241e6e60c4fb927cd1020aR130) to check this earlier -- let me know if that's problematic.)
- Wherever possible, I've added `from __future__ import unicode_literals` so that the distinction between bytestring literals and unicode literals works the same on Python 2.x and 3.x -- hopefully this makes it clearer where we're translating between the two.

The test suite is currently passing, but I want to emphasise two caveats at this point:

1. I *have not yet tested this* properly -- i.e. hooking it up to NSQ and actually trying to send serious amounts of data through it. I wouldn't want to suggest you merge it until I (or someone, at least) has had a chance to do that.

2. I'm not currently clear on whether [this patch](https://github.com/ijl/pynsq/commit/ff25c744a28e66c046ac52c299d6954c2e9f83af) is still necessary. Perhaps I'll find that out when I try and use this in anger.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - 2.6
   - 2.7
+  - 3.5
 env:
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=3.2
   - NSQ_DOWNLOAD=nsq-0.3.0.linux-amd64.go1.3.3 TORNADO_VERSION=4.1
--- a/nsq/_compat.py
+++ b/nsq/_compat.py
@@ -15,6 +15,9 @@ if not PY2:
     text_type = str
     string_types = (str,)
 
+    itervalues = lambda d, *args, **kwargs: d.values(*args, **kwargs)
+    iteritems = lambda d, *args, **kwargs: d.items(*args, **kwargs)
+
     def to_bytes(x, charset='utf-8', errors='strict'):
         if isinstance(x, (bytes, bytearray, memoryview)):
             return bytes(x)
@@ -25,6 +28,9 @@ if not PY2:
 else:
     text_type = unicode
     string_types = (str, unicode)
+
+    itervalues = lambda d, *args, **kwargs: d.itervalues(*args, **kwargs)
+    iteritems = lambda d, *args, **kwargs: d.iteritems(*args, **kwargs)
 
     def to_bytes(x, charset='utf-8', errors='strict'):
         if isinstance(x, (bytes, bytearray, buffer)):
--- a/nsq/_compat.py
+++ b/nsq/_compat.py
@@ -14,6 +14,7 @@ PY2 = sys.version_info[0] == 2
 if not PY2:
     text_type = str
     string_types = (str,)
+    integer_types = (int,)
 
     itervalues = lambda d, *args, **kwargs: d.values(*args, **kwargs)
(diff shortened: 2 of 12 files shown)


nickstenning:
Oh, and as mentioned in #107, this is currently targeting Python 3.5 -- older versions of 3.x are progressively more of a pain, so I'd suggest waiting until somebody complains before doing the work to support them...

mreiferson:
> Oh, and as mentioned in #107, this is currently targeting Python 3.5 -- older versions of 3.x are progressively more of a pain, so I'd suggest waiting until somebody complains before doing the work to support them...

@nickstenning assuming you have the bandwidth to give it a sanity test in an actual cluster, we can merge and stamp and alpha and ask for users to give it a spin. But yes, before stamping an actual release we should do some real testing :grimacing: :fire:

mreiferson:
Also, nice work, appreciate the attention to detail here!

santagada:
Can this be merged?

mreiferson:
I think it just needs to be validated

santagada:
I'm running it fine, so it is a :+1: from me
the docs needs to be updated as well but that can be done after

mreiferson:
I guess I should spin this up on 2.7 and see what happens

santagada:
I think the .travis should be adjusted to run only go 1.4.3, latest 1.5 and latest 1.6 (not 1.3 like it is right now) and test only on python 2.7 and 3.4+ (no need for python 2.6 anymore). also tornado is at version 4.3 now.

mreiferson:
@santagada I'm going to find some time soon to experiment with this on 2.7 before landing, stay tuned!

misja:
Just a short note, am running in production with this set of diffs, including a missed change https://github.com/nickstenning/pynsq/pull/2. ~200/sec, ~40 mill processed already using the asyncio event loop, non-stop. Looks good, would love to see this get merged in main :)

nickstenning:
Thanks, @misja. I've pulled your change from 400ece3 into 1c8d78b. This branch has been rebased on latest master.

mreiferson:
thanks @nickstenning and @misja - still haven't had a chance to run this on 2.7 but I appreciate everyone giving this a run!

mreiferson:
OK - giving this a run on 2.7, will report back in a few days 🔥 👊

mreiferson:
been running this for over 24hrs without issues on Python 2.7

LGTM

thanks everyone!

nickstenning:
🎉 Thanks for the review and merge @mreiferson!

misja:
yay, thanks all :)

PR #157, opened 2016-06-05
Reuse struct objects

virtuald:
From the python docs: 

> Creating a Struct object once and calling its methods is more efficient than calling the struct functions with the same format since the format string only needs to be compiled once.

### DIFF
--- a/nsq/_compat.py
+++ b/nsq/_compat.py
@@ -28,11 +28,8 @@ if not PY2:
             return x.encode(charset, errors)
         raise TypeError('expected bytes or a string, not %r' % type(x))
 
-    def struct_pack(fmt, *values):
-        return struct.pack(fmt, *values)
-
-    def struct_unpack(fmt, string):
-        return struct.unpack(fmt, string)
+    def _create_struct(fmt):
+        return struct.Struct(fmt)
 
 else:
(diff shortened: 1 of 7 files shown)


mreiferson:
@virtuald thanks!  This one looks fine to me, I'm curious if you recorded any statistically significant change in performance from this?

virtuald:
Yes, it was on the list of "things towards the top of the profiling output" (I think like 1-2% percent time) to practically 0% time with this patch. I'll need to re-run it on my other computer to get exact data if you want it.

mreiferson:
SGTM

Would you mind pasting that profiling output?

virtuald:
Sorry for the delay, here's what I've got using my rawperf script -- didn't profile publishing, just reading:

* Original 40a637b02c28edfb7723373b81f7e7b9d5e364ed: 3 runs of 13.543s/13.636s/13.391s
  * profiled: time was 19.171s, struct unpack related percentages were 1.12% + 0.00% + 0.46%

* Patched a0628262936d25228e1bf8c06297627f7214b992: 3 runs of 13.244s/13.078s/13.048s
  * profiled: time was 18.955s, struct unpack related percentages were 0.23% + 0.15% + 0.07%

PR #158, opened 2016-06-06
Minor performance tweaks

virtuald:
The first one is pretty self-explanatory (maybe needs another PR?) -- only call `time.time()` once, the extra function overhead (despite how small it is) isn't necessary.

The second one we can debate, in the reader it now immediately calls _start_read() instead of going through the IOLoop to call it. This results in a ~1s improvement in my [rawperf script](https://gist.github.com/virtuald/d7416e06d0d348af30aa1425fb67a5b6). I think I understand why it was done this way originally -- presumably don't want to starve the IOLoop if messages just happen to come in at the perfect rate where the socket buffer can keep feeding messages... since the read functions will call each other continuously in a loop without yielding to the IOLoop.

But... I wonder if that can actually happen?

If it can happen... then perhaps it should only use add_callback every once in awhile, or if it detects a cycle... or some other probably terrible thing? At the very least, should add a note as to why the code is using add_callback here.

### DIFF
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -277,7 +277,7 @@ class AsyncConn(event.EventedMixin):
             self.trigger(event.DATA, conn=self, data=data)
         except Exception:
             logger.exception('uncaught exception in data event')
-        self.io_loop.add_callback(self._start_read)
+        self._start_read()
 
     def send(self, data):
         self.stream.write(data)
--- a/tests/test_async.py
+++ b/tests/test_async.py
@@ -100,17 +100,15 @@ def test_read_body():
     on_data = create_autospec(f)
     conn.on('data', on_data)
 
-    mock_ioloop_addcb = create_autospec(f)
-    mock_io_loop.add_callback = mock_ioloop_addcb
     data = 'NSQ'
     conn._read_body(data)
     on_data.assert_called_once_with(conn=conn, data=data)
-    mock_ioloop_addcb.assert_called_once_with(conn._start_read)
+    conn.stream.read_bytes.assert_called_once_with(4, conn._read_size)
 
     # now test functionality when the data_callback fails
     on_data.reset_mock()
-    mock_ioloop_addcb.reset_mock()
+    conn.stream.read_bytes.reset_mock()
     on_data.return_value = Exception("Boom.")
     conn._read_body(data)
-    # verify that we still added callback for the next start_read
-    mock_ioloop_addcb.assert_called_once_with(conn._start_read)
+    # verify that the next _start_read was called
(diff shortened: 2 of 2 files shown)


mreiferson:
@virtuald same question here - any statistically significant improvement here?

I worry about chasing micro-optimizations...

mreiferson:
Above question was w/r/t the `time.time` change, sorry.

re: the deferred read, I think the change is safe because `_read_bytes` just calls `self.stream.read_bytes`, which defers if necessary.

👍

virtuald:
The time.time change isn't significant.

Heh, well, it's clear that whomever wrote that code originally really wanted to make sure that it defers to the ioloop, since the test checks for it. I'll fix the test later today.

nsq/async.py
```diff
@@ -277,7 +277,7 @@ def _read_body(self, data):
             self.trigger(event.DATA, conn=self, data=data)
         except Exception:
             logger.exception('uncaught exception in data event')
-        self.io_loop.add_callback(self._start_read)
```
jehiah: I have several spots where i use several independent reader's on the same IOLoop so yielding is important to me. It's also important because calling things directly can lead to infinite call stacks. I think that makes me :thumbsdown: for this change.

virtuald: Yeah, that seems reasonable enough. I wonder if there's a more efficient way of doing this then?

mreiferson: I'm pretty sure it won't "block", it'll either immediately handle unprocessed (but read) bytes or it will defer and setup a callback.

http://www.tornadoweb.org/en/stable/_modules/tornado/iostream.html#BaseIOStream.read_bytes

This just skips an _additional_ deferment.

mreiferson:
Let's drop the `time.time` change if you don't mind.

Assuming @jehiah is amenable to the other change and we fix the failing test, this LGTM!

nsq/async.py
virtuald: Right -- the question I have is whether it's possible that enough data could be fed in at the right rhythm that would cause it to never defer. That's something I'm a bit unsure about.

mreiferson: From my scanning of the tornado code, I think it's functionally equivalent with one less deferment.

jehiah: @mreiferson thanks for digging a little further. I agree, and that addresses my previous concern.

virtuald:
I fixed the tests... it looks like one of the tests on Travis failed with a spurious error (connection refused to nsqd), not totally sure what caused that.

mreiferson:
restarted those individual tests, let's see if we get a 🍏

mreiferson:
LGTM

mreiferson:
thanks!

PR #161, opened 2016-08-31
fix protocol.SendError in python 3.4

xiaost:
% formatting is availble since python 3.5

see: http://legacy.python.org/dev/peps/pep-0461/


mreiferson:
LGTM, thanks!

PR #167, opened 2017-02-17
1.0 compatible lookupd queries

jehiah:
pynsq was not yet making 1.0 compatible queries to lookupd, so this code breaks due to the deprication of the old format in nsqio/nsq#367. This updates to pass the right header to make forwards compatible queries.

### DIFF
--- a/nsq/reader.py
+++ b/nsq/reader.py
@@ -582,6 +582,7 @@ class Reader(Client):
 
         req = tornado.httpclient.HTTPRequest(
             lookupd_url, method='GET',
+            header={"Accept":"application/vnd.nsq; version=1.0"},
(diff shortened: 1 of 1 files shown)


mreiferson:
nice

nsq/reader.py
```diff
@@ -582,6 +582,7 @@ def query_lookupd(self):
 
         req = tornado.httpclient.HTTPRequest(
             lookupd_url, method='GET',
+            header={"Accept":"application/vnd.nsq; version=1.0"},
```
jehiah: typo! should be `headers`. 😢

mreiferson: hahahahah 😭

PR #168, opened 2017-02-17
Fix nsqlookupd 1.0 compatible queries

jehiah:
This fixes a regression from #167.

clearly @mreiferson i should validate this before we merge =)


mreiferson:
is this ready?

jehiah:
:thumbsup: finished validating this.

PR #169, opened 2017-02-21
IDENTIFY: send 1.0 compatible client_id and hostname fields

jehiah:
This adds support for new IDENTIFY fields as support for the old fields was removed from nsqd in nsqio/nsq#367


mreiferson:
we're good at computers

PR #170, opened 2017-03-27
Use TCP_NODELAY for improved latency

protoss-player:
According to the NSQ protocol, consumers have to send `FIN`s for every
message they receive, which makes them extremely slow with low `max_in_flight`
values due to TCP buffering.

Adding TCP_NODELAY option on `async.AsyncConn`'s sockets makes `nsq.Reader`
substantially faster. My tests show increase from ~50 to ~4.5K messages per
second on localhost with `max_in_flight = 1`.

----

NODELAY will be enabled by default. It may be disabled by passing `nodelay=False` to `Reader`/`Writer` ctor's kwargs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@ dist
 pynsq.egg-info
 build
 docs/_build
+.cache
+.eggs
+*.dat
+*.tmp
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -225,6 +225,7 @@ class AsyncConn(event.EventedMixin):
 
         self.stream = tornado.iostream.IOStream(self.socket, io_loop=self.io_loop)
         self.stream.set_close_callback(self._socket_close)
+        self.stream.set_nodelay(True)
 
         self.state = CONNECTING
         self.on(event.CONNECT, self._on_connect)


ploxiln:
Do you think there is ever real downside to enabling nodelay for nsq tcp connections? (Is it ever common for pynsq to do a sequence of small writes that go out as separate packets without nodelay?) Perhaps, instead of adding a performance tuning-option, it should just be enabled?

mreiferson:
@ploxiln I was thinking the same (and was actually surprised this wasn't already the case)

👍 to on by default

protoss-player:
@ploxiln I don't think NODELAY could break anything, especially when communicating over LAN. BTW, it's enabled by default for all TCP sockets in Asyncio, see https://github.com/python/cpython/blob/master/Lib/asyncio/selector_events.py#L689.

.gitignore
```diff
@@ -3,3 +3,7 @@ dist
 pynsq.egg-info
 build
 docs/_build
+.cache
+.eggs
+*.dat
+*.tmp
```
ploxiln: this change is not really related to the rest

ploxiln:
To clarify, in addition to "on by default", I was proposing not adding any option to disable it.

.gitignore
protoss-player: Yep, indeed, but running tests produced too much crap. Not sure if this could make a decent PR on its own.

nsq/async.py
```diff
@@ -128,6 +128,8 @@ class AsyncConn(event.EventedMixin):
     :param msg_timeout: the amount of time (in seconds) that nsqd will wait
         before considering messages that have been delivered to this
         consumer timed out (requires nsqd 0.2.28+)
+
+    :param nodelay: set TCP NODELAY mode on socket
```
mreiferson: as per @ploxiln's suggestion, let's drop the option altogether!

protoss-player: OK, done

mreiferson:
@protoss-player would you mind squashing down to one commit please?

protoss-player:
@mreiferson Done.

BTW, I believe there's a "squash merge" option for PRs (some changes to repository settings may be required), see https://github.com/blog/2141-squash-your-commits. Never used it myself though.

mreiferson:
I'm going to merge this, but it looks like our travis bootstrapping is broken and all tests are failing.

I'll fix separately.

Thanks for the contribution!

PR #174, opened 2017-04-28
writer: discard closed connections in pub

panaetov:
Writer publishes messages on opened connections only.

### DIFF
--- a/nsq/writer.py
+++ b/nsq/writer.py
@@ -128,11 +128,15 @@ class Writer(Client):
             callback = functools.partial(self._finish_pub, command=command,
                                          topic=topic, msg=msg)
 
-        if not self.conns:
-            callback(None, protocol.SendError('no connections'))
+        open_connections = [
+            conn for conn in self.conns.values()
+            if conn.connected()
+        ]
+        if not open_connections:
+            callback(None, protocol.SendError('no open connections'))
             return
 
-        conn = random.choice(list(self.conns.values()))
+        conn = random.choice(open_connections)
         conn.callback_queue.append(callback)
         cmd = getattr(protocol, command)
 
@@ -145,6 +149,7 @@ class Writer(Client):
             conn.send(cmd(*args))
         except Exception:
             logger.exception('[%s] failed to send %s' % (conn.id, command))
+            callback(None, protocol.SendError('send error'))
             conn.close()
 
     def _on_connection_error(self, conn, error, **kwargs):
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -215,8 +215,17 @@ class AsyncConn(event.EventedMixin):
     def __str__(self):
         return self.host + ':' + str(self.port)
 
+    def connected(self):
+        return self.state == CONNECTED
+
+    def connecting(self):
+        return self.state == CONNECTING
+
+    def closed(self):
+        return self.state in (INIT, DISCONNECTED)
+
     def connect(self):
-        if self.state not in [INIT, DISCONNECTED]:
+        if not self.closed():
             return
 
         self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)


nsq/writer.py
```diff
@@ -128,11 +128,14 @@ def _pub(self, command, topic, msg, delay_ms=None, callback=None):
...
-            callback(None, protocol.SendError('no connections'))
-            return
-
-        conn = random.choice(list(self.conns.values()))
+        open_connects = [
+            conn for conn in self.conns.values()
+            if conn.connected()
+        ]
+        if not open_connects:
+            raise PubFailure("Not open connections")
```
mreiferson: I don't want to introduce different failure semantics (exceptions) — let's use:

```python
callback(None, protocol.SendError('no open connections'))
```

nsq/writer.py
```diff
@@ -128,11 +128,14 @@ def _pub(self, command, topic, msg, delay_ms=None, callback=None):
             callback = functools.partial(self._finish_pub, command=command,
                                          topic=topic, msg=msg)
 
-        if not self.conns:
-            callback(None, protocol.SendError('no connections'))
-            return
-
-        conn = random.choice(list(self.conns.values()))
+        open_connects = [
```
mreiferson: `s/open_connects/open_connections`

nsq/writer.py
```diff
@@ -143,9 +146,9 @@ def _pub(self, command, topic, msg, delay_ms=None, callback=None):
 
         try:
             conn.send(cmd(*args))
-        except Exception:
-            logger.exception('[%s] failed to send %s' % (conn.id, command))
+        except:
             conn.close()
+            raise PubFailure('[%s] failed to send %s' % (conn.id, command))
```
mreiferson: same here, don't want to introduce new failure semantics, let's drop the `raise` please

nsq/writer.py
panaetov: Done

nsq/writer.py
panaetov: Done

nsq/writer.py
panaetov: Done

nsq/writer.py
```diff
@@ -143,8 +147,9 @@ def _pub(self, command, topic, msg, delay_ms=None, callback=None):
 
         try:
             conn.send(cmd(*args))
-        except Exception:
+        except:
```
ploxiln: Why this change?

It's generally a good idea to catch no more broadly than `Exception`. There are a couple exceptions which intentionally do not subclass it, like `SystemExit` and `KeyboardInterrupt`, because catching those can (intermittently) break exiting and restarting the process.

panaetov: Thanks, done.

nsq/writer.py
```diff
@@ -145,6 +149,7 @@ def _pub(self, command, topic, msg, delay_ms=None, callback=None):
             conn.send(cmd(*args))
         except Exception:
             logger.exception('[%s] failed to send %s' % (conn.id, command))
+            callback(None, protocol.SendError('sending error'))
```
mreiferson: I'm not 100% sure of this change — what issue is this addressing?

ploxiln: I think that without this, the caller is not informed that it possibly failed to send the message (even after seemingly having a connection).

mreiferson: Ahh, right. In that case `s/sending error/send error` please :)

panaetov: Done

mreiferson:
LGTM, thanks for sticking with this! Would you mind squashing down to one commit?

protoss-player:
Hi @mreiferson,

I believe you can squash everything down when merging this PR. As per https://github.com/blog/2141-squash-your-commits, there should be "Confirm squash and merge" option.

mreiferson:
Yea, I know, I actually prefer the merge commit for historical reasons, and that option isn't offered in GH.

panaetov:
Done. Everything is squashed down to one commit.

mreiferson:
Thanks!

Issue #177, opened 2017-05-05
reader: max_in_flight check in _send_rdy() doesn't take in-flight msgs into account


alpaker:
`Reader._send_rdy()` contains logic to avoid setting a `RDY` count on a connnection that would allow exceeding the reader's `max_in_flight`.  The [code](https://github.com/nsqio/pynsq/blob/master/nsq/reader.py#L455) in question is:
```
if (self.total_rdy + value) > self.max_in_flight
    if not conn.rdy:
        # if we're going from RDY 0 to non-0 and we couldn't because
        # of the configured max in flight, try again
        rdy_retry_callback = functools.partial(self._rdy_retry, conn, value)
        conn.rdy_timeout = self.io_loop.add_timeout(time.time() + 5, rdy_retry_callback)
    return
```
The condition `(self.total_rdy + value) > self.max_in_flight` doesn't take into account messages that are currently in flight.  This creates an opportunity for `max_in_flight` violations.  The recipe in [this report](https://github.com/nsqio/pynsq/issues/175) will produce one, usually within a minute.

I reported this offline to @jonmorehouse, who pointed out that it was surprising the issue hasn't surfaced before.  I think the reason is that it can only hit when `max_in_flight` is less than the connection count, and this use case must be rare in practice:

* `_send_rdy()` is only called with a `value` parameter of 0, 1 or `_connection_max_in_flight()`.
* So if `len(conns) * _connection_max_in_flight() <= max_in_flight`, then we'll never violate the in-flight constraint, because we're implicitly restricting ourselves to a per-connection limit that guarantees safety.
* But if `len(conns) > max_in_flight` and ` _connection_max_in_flight() == 1`, then respecting the per-connection limit isn't enough.  In this case deciding whether a non-zero `RDY` count is permissible on a connection requires taking into account the in-flight count on other connections.

alpaker:
The same issue affects the Go client.

An appropriate fix for the Python client would be:
```
diff --git a/nsq/reader.py b/nsq/reader.py
index ad6447a..4ad1281 100644
--- a/nsq/reader.py
+++ b/nsq/reader.py
@@ -452,7 +452,10 @@ class Reader(Client):
         if value > conn.max_rdy_count:
             value = conn.max_rdy_count

-        if (self.total_rdy + value) > self.max_in_flight:
+        total_possible_other = sum(
+            max(c.last_rdy, c.in_flight) for c in self.conns.values() if c is not conn
+        )
[...]

PR #179, opened 2017-05-10
reader:  don't decrement total_rdy on message receipt

alpaker:
This brings reader behavior into agreement with nsqd behavior (after nsqio/nsq#404) and removes an opportunity for max_in_flight violations (#177).

This [block](https://github.com/nsqio/pynsq/blob/master/nsq/reader.py#L322-L332) was removed because it's redundant given `_redistributed_rdy()` and it's easier to maintain invariants without it.

@mreiferson

### DIFF
--- a/nsq/reader.py
+++ b/nsq/reader.py
@@ -317,21 +317,7 @@ class Reader(Client):
             logger.exception('[%s:%s] failed to handle_message() %r', conn.id, self.name, message)
 
     def _handle_message(self, conn, message):
-        self.total_rdy = max(self.total_rdy - 1, 0)
-
-        rdy_conn = conn
-        if len(self.conns) > self.max_in_flight and time.time() - self.random_rdy_ts > 30:
-            # if all connections aren't getting RDY
-            # occsionally randomize which connection gets RDY
-            self.random_rdy_ts = time.time()
-            conns_with_no_rdy = [c for c in itervalues(self.conns) if not c.rdy]
-            if conns_with_no_rdy:
-                rdy_conn = random.choice(conns_with_no_rdy)
-                if rdy_conn is not conn:
-                    logger.info('[%s:%s] redistributing RDY to %s',
-                                conn.id, self.name, rdy_conn.id)
-
-        self._maybe_update_rdy(rdy_conn)
+        self._maybe_update_rdy(conn)
 
         success = False
         try:
@@ -358,7 +344,9 @@ class Reader(Client):
         if self.backoff_timer.get_interval() or self.max_in_flight == 0:
             return
 
-        if conn.rdy <= 1 or conn.rdy < int(conn.last_rdy * 0.25):
+        # On a new connection or in backoff we start with a tentative RDY count
+        # of 1.  After successfully receiving a first message we go to full throttle.
+        if conn.rdy == 1:
             self._send_rdy(conn, self._connection_max_in_flight())
 
     def _finish_backoff_block(self):
@@ -452,15 +440,10 @@ class Reader(Client):
         if value > conn.max_rdy_count:
             value = conn.max_rdy_count
 
-        if (self.total_rdy + value) > self.max_in_flight:
-            if not conn.rdy:
-                # if we're going from RDY 0 to non-0 and we couldn't because
-                # of the configured max in flight, try again
-                rdy_retry_callback = functools.partial(self._rdy_retry, conn, value)
-                conn.rdy_timeout = self.io_loop.add_timeout(time.time() + 5, rdy_retry_callback)
+        new_rdy = max(self.total_rdy - conn.rdy + value, 0)
+        if new_rdy > self.max_in_flight:
             return
 
-        new_rdy = max(self.total_rdy - conn.rdy + value, 0)
         if conn.send_rdy(value):
             self.total_rdy = new_rdy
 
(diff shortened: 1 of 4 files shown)


mreiferson:
Thanks! I'm gonna read through the diff more carefully, but in the meantime tests are failing on Python 3.5 https://travis-ci.org/nsqio/pynsq/jobs/230827968

alpaker:
D'oh!  Fixed.

nsq/reader.py
```diff
@@ -358,7 +344,7 @@ def _maybe_update_rdy(self, conn):
         if self.backoff_timer.get_interval() or self.max_in_flight == 0:
             return
 
-        if conn.rdy <= 1 or conn.rdy < int(conn.last_rdy * 0.25):
+        if conn.rdy == 1:
```
mreiferson: I think I understand this change but want to make sure — this is because we initially set `RDY 1` and we want to make sure we adjust to an appropriate per-connection max-in-flight, right?

mreiferson: Perhaps it's worth a comment since it took me a little while to figure this out.

nsq/reader.py
```diff
@@ -665,10 +646,21 @@ def _redistribute_rdy_state(self):
                     logger.info('[%s:%s] idle connection, giving up RDY count', conn.id, self.name)
                     self._send_rdy(conn, 0)
 
+            conns = self.conns.values()
+
+            in_flight = [c for c in conns if c.in_flight]
```
mreiferson: I still don't quite understand why we need to use `in_flight` when `total_rdy` should represent the same thing? The accounting for both `total_rdy` and `in_flight` used to happen [in 2 subsequent lines in the same block](https://github.com/nsqio/pynsq/pull/179/files#diff-9fd12dcd9000f610741f4d399c31e1b8R489)).

nsq/reader.py
```diff
@@ -677,7 +669,7 @@ def _redistribute_rdy_state(self):
             # We also don't attempt to avoid the connections who previously might have had RDY 1
             # because it would be overly complicated and not actually worth it (ie. given enough
             # redistribution rounds it doesn't matter).
-            possible_conns = list(self.conns.values())
+            possible_conns = [c for c in conns if not (c.in_flight or c.rdy)]
```
mreiferson: Same comment here re: use of `in_flight`

nsq/reader.py
mreiferson: This might be further complicated because the `max_in_flight` variable below is a terrible name. Perhaps a better name would be `available_rdy`?

nsq/reader.py
alpaker: Right.  Updating `RDY` here used to serve two purposes:  Periodically reupping `RDY` when it dipped too low, and going from an initial throttled `RDY 1` on startup to full-throttle connection-max-in-flight.  Now only the second case is relevant.

nsq/reader.py
alpaker: Consider a case where `max_in_flight` is less than the connection count and a task takes longer than `low_rdy_idle_timeout`.   Then we'll set `RDY 0` on a connection even while it's still processing a message (https://github.com/nsqio/pynsq/blob/master/nsq/reader.py#L664).  So we can end up in a situation where `total_rdy` is strictly less than the total count of in-flight messages.  If we go from from `RDY 0` to `RDY 1` on a connection that has messages available for delivery, we guarantee a max-in-flight violation.

alpaker: The above last sentence should be: 

If we go from from `RDY 0` to `RDY 1` on a connection (other than the one just set to `RDY 0`) that has messages available for delivery, we guarantee a max-in-flight violation.

alpaker: Note that this is not difficult to trigger.  All you need is:

1. `max_in_flight` less than the number of connections

mreiferson: OK, I think that makes sense, but what about the section above? The sum of `in_flight` can also be less than `total_rdy`, in which case we'd enter this loop below thinking we have more than we actually do to give out?

alpaker: Sorry, I think I misunderstood our earlier discussion to have meant that that was the kind of case you weren't worried about.  The possibility you mention would be taken care of with:
```
            in_flight_or_rdy = sum(max(c.in_flight, c.rdy) for c in conns)
            if backoff_interval:
                max_in_flight = max(0, 1 - in_flight_or_rdy)
            else:
                max_in_flight = self.max_in_flight - in_flight_or_rdy
```
That's essentially the logic I proposed in https://github.com/nsqio/pynsq/issues/177#issuecomment-299917567.

alpaker: With the proposed changes, the full diff would look like https://github.com/nsqio/pynsq/compare/master...alpaker:no-decr-total-rdy-2?diff=unified&expand=1&name=no-decr-total-rdy-2#diff-2a8bc85bf9c95f482da1eb0490a60251R654

mreiferson: Cool, let's update this PR with that change.

As per above, I also think we should prefix the `max_in_flight` variable in this method with `available_`.

Thanks!

alpaker:
@mreiferson Updated as discussed.  (I went with your earlier suggestion of `available_rdy` for the variable name in `_redistribute_rdy()`.)

nsq/reader.py
```diff
@@ -665,10 +648,27 @@ def _redistribute_rdy_state(self):
...
+            # round can proceed and we don't stay pinned to the same connections.
+            #
+            # if nothing's in flight, then we have connections with RDY 1 that are still
+            # waiting to hit the idle timeout, in which case it's ok to do nothing.
+            if not available_rdy:
+                try:
+                    c = random.choice([c for c in conns if c.in_flight])
+                    logger.info('[%s:%s] too many msgs in flight, giving up RDY count', c.id, self.name)
+                    self._send_rdy(c, 0)
+                except IndexError:
```
mreiferson: Prefer not using exceptions for flow control here.

nsq/reader.py
```diff
@@ -665,10 +648,27 @@ def _redistribute_rdy_state(self):
...
 
+            conns = self.conns.values()
+
+            in_flight_or_rdy = len([c for c in conns if c.in_flight or c.rdy])
             if backoff_interval:
-                max_in_flight = 1 - self.total_rdy
+                available_rdy = max(0, 1 - in_flight_or_rdy)
             else:
-                max_in_flight = self.max_in_flight - self.total_rdy
+                available_rdy = self.max_in_flight - in_flight_or_rdy
```
mreiferson: Now that we're using `in_flight` _and_ `rdy`, I think this needs a `max(0, ...)` too

alpaker: Both issues fixed.

PR #183, opened 2017-06-05
reader: make sure RDY is updated when connection count changes

### LINKED ISSUE #182: pynsq 0.8.1: Wrong distribution of ready count
Software version:
pynsq 0.8.1
nsq v1.0.0-compat

I am running NSQ cluster with the following configuration (1 lookupd + 2 nsqd):

```
Running nsqlookupd on ports: tcp=10000; http=10001
Running nsqd on ports: tcp=10100; http=10101
Running nsqd on ports: tcp=10200; http=10201

```

Then I am running the following code (it writes one message, then waiting for this message to read back, then writes another message and so on).
And it is hanging forever after a few round tripped messages:

```python
import nsq
from tornado import ioloop
from tornado.queues import Queue
from tornado import gen
from uuid import uuid4
import time


class Writer(nsq.Writer):

    def __init__(self, *args, **kwargs):
        super(Writer, self).__init__(*args, **kwargs)
        self._connected_future = gen.Future()

    def push(self, topic, message):
        future = gen.Future()

        def callback(conn, result):
            if result == 'OK':
                future.set_result(result)
            else:
                future.set_exception(result)

        self.pub(topic, message, callback=callback)
        return future

    def _on_connection_ready(self, conn, **kwargs):
        super(Writer, self)._on_connection_ready(conn, **kwargs)

        if not self._connected_future.done():
            self._connected_future.set_result(None)

    def wait_for_connection(self):
        return self._connected_future


class Reader(nsq.Reader):

    def __init__(self, *args, **kwargs):
        super(Reader, self).__init__(*args, message_handler=self.__handle_message, **kwargs)
        self._messages = Queue()

    def __handle_message(self, message):
        message.enable_async()
        self._messages.put_nowait(message)

    def pop(self):
        return self._messages.get()


@gen.coroutine
def main():
    channel = topic = 'test#ephemeral'
    nsqd_addresses = [
        'localhost:10100',
        'localhost:10200'
    ]
    reader = Reader(
        topic,
        channel,
        lookupd_http_addresses='http://localhost:10001',
        lookupd_poll_interval=1,
        max_in_flight=32,
        output_buffer_timeout=1
    )
    writer = Writer(nsqd_tcp_addresses=nsqd_addresses)
    yield writer.wait_for_connection()

    for i in xrange(10**6):
        message_text = uuid4().hex
        start_time = time.time()
        yield writer.push(topic, message_text)
        message = yield reader.pop()
        message.finish()
        assert message.body == message_text
        dt = time.time() - start_time
        print '[{}] Round-trip time: {:.02f}ms'.format(i, dt * 1000)


ioloop.IOLoop.current().run_sync(main)

```

The output sample:
```
[0] Round-trip time: 1189.58ms
[1] Round-trip time: 128.57ms
[2] Round-trip time: 0.48ms
[3] Round-trip time: 0.62ms
[4] Round-trip time: 0.95ms
# here it hangs on the 5th message
```

That's what I see in nsqadmin: 

Ready count values between two connections distributed not equally (32 vs 0). While in pynsq 0.8.0 it works fine (16 vs 16):

alpaker:
Fixes #182.

Previously, a connection's `rdy` attribute was decremented on message receipt; when it [dipped low enough](https://github.com/nsqio/pynsq/blob/07a75bd5e99ade9d1f6a13593d5bfa17e67c17f4/nsq/reader.py#L361) the connection would send a new `RDY` count of `_connection_max_in_flight()`.  One consequence of this was that, when `_connection_max_in_flight()` changed (because either the connection count or the reader's `max_in_flight` attribute changed), each connection would eventually update to the new value.

The reader's `total_rdy` attribute was also decremented on message receipt.  This guaranteed (under the assumption that some connection was receiving messages), that `total_rdy` would drop below `max_in_flight`, freeing up space for a new connection to send an initial `RDY` count of 1.

Now that these client-side attributes aren't updated with each message received, these two cases need to be catered for explicitly.

### DIFF
--- a/nsq/reader.py
+++ b/nsq/reader.py
@@ -344,10 +344,15 @@ class Reader(Client):
         if self.backoff_timer.get_interval() or self.max_in_flight == 0:
             return
 
-        # On a new connection or in backoff we start with a tentative RDY count
-        # of 1.  After successfully receiving a first message we go to full throttle.
-        if conn.rdy == 1:
-            self._send_rdy(conn, self._connection_max_in_flight())
+        # Update RDY in 2 cases:
+        #     1. On a new connection or in backoff we start with a tentative RDY
+        #        count of 1.  After successfully receiving a first message we go to
+        #        full throttle.
+        #     2. After a change in connection count or max_in_flight we adjust to the new
+        #        connection_max_in_flight.
+        conn_max_in_flight = self._connection_max_in_flight()
+        if conn.rdy == 1 or conn.rdy != conn_max_in_flight:
+            self._send_rdy(conn, conn_max_in_flight)
 
     def _finish_backoff_block(self):
         self.backoff_timeout = None
@@ -503,6 +508,12 @@ class Reader(Client):
                 conn.id, self.name, conn.max_rdy_count, self.max_in_flight)
 
         self.conns[conn.id] = conn
+
+        conn_max_in_flight = self._connection_max_in_flight()
+        for c in self.conns.values():
+            if c.rdy > conn_max_in_flight:
+                self._send_rdy(c, conn_max_in_flight)
+
         # we send an initial RDY of 1 up to our configured max_in_flight
         # this resolves two cases:
         #    1. `max_in_flight >= num_conns` ensuring that no connections are ever
--- a/tests/reader_unit_test_helpers.py
+++ b/tests/reader_unit_test_helpers.py
@@ -1,0 +1,51 @@
+from __future__ import absolute_import
+from __future__ import with_statement
+
+from mock import patch, create_autospec
+from tornado.ioloop import IOLoop
+
+import nsq
+from nsq import event
+
+_conn_port = 4150
+
+
+def get_reader(io_loop=None, max_in_flight=5):
+    return nsq.Reader("test", "test",
+                      message_handler=_message_handler,
+                      lookupd_http_addresses=["http://test.local:4161"],
+                      max_in_flight=max_in_flight,
+                      io_loop=io_loop)
+
+
+def get_ioloop():
+    ioloop = create_autospec(IOLoop)
+    ioloop.time.return_value = 0
+    return ioloop
+
+
+def get_conn(reader):
+    global _conn_port
+    with patch('nsq.async.tornado.iostream.IOStream', autospec=True):
+        conn = reader.connect_to_nsqd('localhost', _conn_port)
+    _conn_port += 1
+    conn.trigger(event.READY, conn=conn)
+    return conn
+
+
+def send_message(conn):
+    msg = _get_message(conn)
+    conn.in_flight += 1
+    conn.trigger(event.MESSAGE, conn=conn, message=msg)
+    return msg
+
+
+def _get_message(conn):
+    msg = nsq.Message("1234", "{}", 1234, 0)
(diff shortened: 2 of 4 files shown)


alpaker:
@mreiferson RFR

nsq/reader.py
```diff
@@ -513,6 +519,11 @@ def _on_connection_ready(self, conn, **kwargs):
             # only send RDY 1 if we're not in backoff (some other conn
             # should be testing the waters)
             # (but always send it if we're the first)
+            if preexisting_conns and self.total_rdy == self.max_in_flight:
+                # Preexisting connections might be using all possible RDY, preventing us
+                # from sending RDY 1 on the new connection.  In that case, we force a
+                # preexisting connection to update RDY to the new, lower connection_max_in_flight.
+                self._maybe_update_rdy(preexisting_conns[0])
```
mreiferson: Is it safe to always choose the 0th conn?

mreiferson:
Thinking out loud — is it easier to just reset _all_ conn RDY counts whenever we process a new connection? Rather than this odd separation between initial RDY and then proper per-conn max_in_flight after receipt of first message?

nsq/reader.py
alpaker: I think you're right that it's not safe.  We need to guarantee that the conn we select has `RDY` greater than the new `connection_max_in_flight()`.  One can be shown to exist at this point, on the assumption that we're not in the abnormal regime where the connection count, counting the new conn, is greater than `max_in_flight`.  If we are in the abnormal regime, trying to set a non-zero `RDY` on the new conn is going to fail and we have to rely on periodic redistribution to give `RDY` to the new conn.

Just resetting `RDY` on all the connections, as you suggest, would simplify the logic.

alpaker:
I did try that out.  But test expectations need to be changed if we do that, so instead I went with a change that involved less perturbation to current behavior.  Happy to go either way.

alpaker:
I think we don't want to unconditionally reset RDY counts on the other connections: some might be be testing the waters with RDY 1 and waiting for a message, so it would be wrong to update them to the connection max-in-flight.  But it would work to update RDY on all connections that have RDY greater than the new per-connection max.  With that additional condition, we also don't need any test changes.  PR updated accordingly.

mreiferson:
Sorry for the delay, I think this looks good.

Is there a relatively easy test we can add that fails on `master` and is fixed by this?

alpaker:
It's straightforward to test, but requires multiple connections, so there's not a natural place to add the test.  One option would be to refactor the integration test harness in `test_reader.py` to work with multiple connections.  But a better direction might be to add a new `test_rdy.py` that can accumulate unit tests related to `RDY` count management, à la `test_backoff.py`.

mreiferson:
> But a better direction might be to add a new test_rdy.py that can accumulate unit tests related to RDY count management, à la test_backoff.py.

Let's try that! Thanks!

alpaker:
@mreiferson Added the new test file.  Sorry for the delay.  

I moved some helper functions from `test_backoff.rdy` to a new `reader_unit_test_helpers.py`.  Let me know if you don't like that refactoring.

mreiferson:
awesome, thanks!

smant:
What about 0.8.2 release with fixed #182 ?

PR #184, opened 2017-06-06
Fix converting zero delay to time_ms in Message.requeue()

checkraise:
Before this fix zero delay was not correctly converted and time_ms took its default value -1.


mreiferson:
Thanks!

PR #188, opened 2017-08-02
add support for custom hostname

andyxning:
pynsq now only supports override `user_agent` and not supporting override `hostname`.

In cases we run a consumer in a container, the `hostname` shell command will return a docker id which is a random string. When we check the client info in nsqadmin we can not get any useful info about the host name where consumer running.

With enable override `hostname`, we can set the consumer, running in a container, `hostname` to the host name where the container is running which is helpful when we want to determine which host to get the consumers.

/cc @mreiferson @alpaker

### DIFF
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -128,6 +128,9 @@ class AsyncConn(event.EventedMixin):
     :param msg_timeout: the amount of time (in seconds) that nsqd will wait
         before considering messages that have been delivered to this
         consumer timed out (requires nsqd 0.2.28+)
+
+    :param hostname: a string identifying the host where this client runs
+        (default: ``<hostname>``)
     """
     def __init__(
             self,
@@ -147,7 +150,8 @@ class AsyncConn(event.EventedMixin):
             sample_rate=0,
             io_loop=None,
             auth_secret=None,
-            msg_timeout=None):
+            msg_timeout=None,
+            hostname=None):
         assert isinstance(host, string_types)
         assert isinstance(port, int)
         assert isinstance(timeout, float)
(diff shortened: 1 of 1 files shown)


andyxning:
@mreiferson @alpaker 

[The error](https://travis-ci.org/nsqio/pynsq/jobs/260265310#L555-L808) in travis ci seems nothing to do with the change. Could you please take a look at this?

ploxiln:
Just looking at the test matrix, it seems to be the combo of python-3.5 and tornado-3.2 that fails.

andyxning:
@ploxiln Can you please help to figure out why it is failed.

andyxning:
@ploxiln I am not familiar with tornado. I googled the error and change the test by adding a single IOLoop instance according to [tornado #663 comment](https://github.com/tornadoweb/tornado/issues/663#issuecomment-11998879).

Also according to [this google group thread](https://groups.google.com/d/msg/mongodb-user/ivqfqRgvids/G10Yo2QiG94J), it means that `run_sync` with a coroutine that also calls `run_sync` for `TimeoutError: Operation timed out after None seconds` error.

@mreiferson could you please take a look at this?

ploxiln:
We'll investigate, it'll just be a couple days ...

andyxning:
@ploxiln Thanks. 

BTW, could you also please take a look at [nsq #921](https://github.com/nsqio/nsq/pull/921).

mreiferson:
@andyxning thanks for the contribution!

I reran those broken builds and they worked, which means the issue is transient. From reading the links you've pasted, it looks like we've got some object(s) somewhere that aren't using the correct IOLoop instance.

I'm not particularly concerned with fully investigating this and fixing it "correctly", so I'd rather just leave the test code alone for now. Can you remove those changes and we can land this?

andyxning:
@mreiferson I have removed the changes to test. But it will error again in CI.

PR #189, opened 2017-08-03
Deprecation warnings

misja:
Both `cgi.parse_qs` and `inspect.getargspec` will be deprecated. This pull will replace these with `urllib.parse.parse_qs` and `inspect.signature` respectively if available.

### DIFF
--- a/nsq/_compat.py
+++ b/nsq/_compat.py
@@ -59,7 +59,8 @@ struct_h = _create_struct('>h')
 
 try:
     from urllib import parse as urlparse
-    from urllib.parse import urlencode
+    from urllib.parse import urlencode, parse_qs
 except ImportError:
     import urlparse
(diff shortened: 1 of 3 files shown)


mreiferson:
@misja looks good to me, thanks!

misja:
@mreiferson if you intend to create a release including this change, please hold since I just noticed two issues with replacing `inspect.getargspec` by `inspect.signature` -

- the change tries to load `inspect.getargspec` first which will still trigger the deprecated warning
- I naively assumed the return type of `inspect.getargspec` would be equal to `inspect.signature` but it is not and will break things (tests, tests, ...)

I'll add a proper pull second half next week to have this sorted.

mreiferson:
Thanks for validating and giving us the heads up, I'll wait for your PR!

PR #190, opened 2017-08-09
remove dup message constant

andyxning:
Remove duplicate constant `MESSAGE` definition.

/cc @mreiferson @ploxiln


ploxiln:
nice catch, thanks

PR #191, opened 2017-08-09
consumer: snappy/deflate broken in tornado 4.5

dvolodin7:
Tornado 4.5 uses memoryview instead of strings to send data.
Python's deflate compressor cannot process memoryview directly, so it must be explicitly converted to string.

NB. I believe snappy compression has same problem too

### DIFF
--- a/nsq/deflate_socket.py
+++ b/nsq/deflate_socket.py
@@ -39,6 +39,9 @@ class DeflateSocket(object):
         return uncompressed
 
     def send(self, data):
-        chunk = self._compressor.compress(data)
+        if isinstance(data, memoryview):
+            chunk = self._compressor.compress(str(data))
+        else:
+            chunk = self._compressor.compress(data)
         self._socket.send(chunk + self._compressor.flush(zlib.Z_SYNC_FLUSH))
         return len(data)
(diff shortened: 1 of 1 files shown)


nsq/deflate_socket.py
```diff
@@ -39,6 +39,9 @@ def _recv(self, size, method):
         return uncompressed
 
     def send(self, data):
-        chunk = self._compressor.compress(data)
+        if isinstance(data, memoryview):
```
mreiferson: Is this going to fail in older versions of Python?

mreiferson:
@dvolodin7 thanks for the PR, had one minor question.

nsq/deflate_socket.py
dvolodin7: yes, fails in python 2.7

dvolodin7: Must admin that snappy compression has same problem.

dvolodin7: isinstance(data, memoryview) works in Python 2.7

mreiferson:
OK, great, do you wan to update `snappy_socket.py` too?

dvolodin7:
Sure. I'll test on stand and submit full patch

mreiferson:
fixed in #195

jehiah:
fix by #195

PR #192, opened 2017-08-16
Add separate method to derive function signature args

misja:
As promised, a fix for dealing with `inspect.getargspec` being deprecated and `inspect.signature`'s return type not being compatible. This introduces a separate method `func_args` for deriving the method's signature arguments which wraps either `inspect.signature` or `inspect.getargspec` depending on which is available.

### DIFF
--- a/nsq/_compat.py
+++ b/nsq/_compat.py
@@ -66,6 +66,13 @@ except ImportError:
     from cgi import parse_qs
 
 try:
-    from inspect import getargspec as signature
-except ImportError:
     from inspect import signature
+    def func_args(func):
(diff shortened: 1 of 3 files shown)


misja:
The failing check appears to be a tornado `TimeoutError`, so not related to this pull?

ploxiln:
I restarted the one failing test config and it passed.

We should make tests less flaky, sometime soon 😬  (though sometimes the answer is just that "travis is unreasonably slow at times")

mreiferson:
LGTM

mreiferson:
fixes #189

PR #200, opened 2018-02-05
Update the `NSQ Protocol` url

mecforlove:
Hi, I just updated the `NSQ Protocol` url because the old one return 404 status code.


ploxiln:
Thanks!

@mreiferson I think https://pynsq.readthedocs.io/ was last published for v0.7

PR #202, opened 2018-03-08
auth_secret should be a bytestring

ericb:
When a string is passed to `auth_secret` the AUTH command fails due to the body not being in bytes, but if you pass `auth_secret` as bytes, it will fail the string assertion, rendering auth unusable. 

This changes the assertion to check for bytes types, but it could easily be flipped around to accept a string and auto-convert to bytes.

### DIFF
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -5,7 +5,7 @@ import time
 import socket
 import logging
 
-from ._compat import string_types
+from ._compat import bytes_types, string_types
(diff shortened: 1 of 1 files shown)


ploxiln:
I guess no one tried to use auth with python3 - nice catch

ploxiln:
Thanks for the report, and sorry we forgot about it for a while. I fixed this slightly differently in #217

Issue #203, opened 2018-03-12
Bug: "Import Error _DEFAULT_CA_CERTS" With tornado 5.0


tomzhu6066:
Since tornado updated to 5.0, the latest version of pynsq has conflict with this new tornado:

```
>>> import nsq
Traceback (most recent call last):
  File "/Users/tom/Virtualenvs/athena/lib/python3.5/site-packages/nsq/async.py", line 31, in <module>
    from tornado.simple_httpclient import _default_ca_certs as default_ca_certs
ImportError: cannot import name '_default_ca_certs'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tom/Virtualenvs/athena/lib/python3.5/site-packages/nsq/__init__.py", line 29, in <module>
    from .async import AsyncConn
  File "/Users/tom/Virtualenvs/athena/lib/python3.5/site-packages/nsq/async.py", line 34, in <module>
    from tornado.simple_httpclient import _DEFAULT_CA_CERTS
ImportError: cannot import name '_DEFAULT_CA_CERTS'
```

This bug causes huge consequences on my application and plz fix this ASAP.
Currently my way to fix this bug is to put these lines in requirements.txt to force tornado use 4.5.3:
```
tornado==4.5.3
pynsq
```

tomzhu6066:
oooops I just found out pynsq-0.8.3 fixed that issue. ORZ.. waiting for this release!

ploxiln:
Yup, still, this is a good additional note for #185

kimixsx:
Encountered this issue + 1, where did you find out pynsq-0.8.3 fixed it?

ploxiln:
the change in pynsq-0.8.3 is to require tornado < 5.0

tomzhu6066:
@ploxiln yeah I checked the changes hahaha...

yxy:
same issue

memotype:
What is the status on this issue. Is there a workaround? I just ran "pip3 install pynsq" and trying to import it I get this error. Seems odd that this issue would still be open 6 months later being as it renders the module completely unusable...

ploxiln:
This module is usable if you use tornado < 5.0
(which would be enforced by https://github.com/nsqio/pynsq/blob/master/setup.py#L35 but there has not yet been a pynsq release with that constraint). I suggest you `pip3 install tornado==4.5.3`

ploxiln:
[...]

PR #206, opened 2018-05-15
Fix missing CA certs

Y3K:
https://github.com/nsqio/pynsq/issues/203

### DIFF
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -27,14 +27,7 @@ except ImportError:
 import tornado.iostream
 import tornado.ioloop
 
-try:
-    from tornado.simple_httpclient import _default_ca_certs as default_ca_certs
-except ImportError:
-    # Tornado < 4
-    from tornado.simple_httpclient import _DEFAULT_CA_CERTS
-
-    def default_ca_certs():
-        return _DEFAULT_CA_CERTS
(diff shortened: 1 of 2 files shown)


Y3K:
I thought about that, but that'd mean stop supporting Python < 2.7.9 | 3.4

If there is no problem with that I can refactor it to use the built in `ssl` module.

ploxiln:
I think dropping support for python < 2.7.9 or < 3.4 is reasonable. But I'll wait for other maintainers to weigh in. By the way, here's what tornado-5.0 has in setup.py:

```python
# Verify that the SSL module has all the modern upgrades. Check for several
# names individually since they were introduced at different versions,
# although they should all be present by Python 3.4 or 2.7.9.
if (not hasattr(ssl, 'SSLContext') or
        not hasattr(ssl, 'create_default_context') or
        not hasattr(ssl, 'match_hostname')):
    raise ImportError("Tornado requires an up-to-date SSL module. This means "
                      "Python 2.7.9+ or 3.4+ (although some distributions have "
                      "backported the necessary changes to older versions).")
```

mreiferson:
I'm 👍 for bumping up to those min Python versions, thanks @Y3K!

Issue #207, opened 2018-06-21
Pynsq fails to send large messages when using Deflate / Snappy compression


SpaTmole:
## The Problem:
During sending relatively big messages with one of the specified compression algorithms (deflate, snappy) server responds with Error (unexpected EOF).

## The cause:
When compression type is specified the connection is being upgraded with custom Socket classes, which have own `send` methods. In those send methods original socket had one attempt to send the whole pack of compressed data without validation of integrity. The custom `send` method returns the original length of uncompressed data despite the actual amount of sent bytes.
https://github.com/nsqio/pynsq/blob/v0.8.2/nsq/snappy_socket.py#L43
https://github.com/nsqio/pynsq/blob/v0.8.2/nsq/deflate_socket.py#L44

## Solution:
Remove compress-on-flight inside the `send` method - replace it with the original `socket.send`.
Add Encoder's classes which will compress with specified algorithms **before** sending data through the socket / `tornado.IOstream`.

ploxiln:
Some projects like to always open an issue along with a pull-request, but we are not like that - if you have a problem and a fix, it can all live in a single pull-request.

closing this: addressed by #208

Thanks for figuring this out!

PR #208, opened 2018-06-21
Fix sending data through custom sockets with compression.

SpaTmole:
Issue #207

### DIFF
--- a/nsq/deflate_socket.py
+++ b/nsq/deflate_socket.py
@@ -7,9 +7,7 @@ import errno
 
 class DeflateSocket(object):
     def __init__(self, socket, level):
-        wbits = -zlib.MAX_WBITS
-        self._decompressor = zlib.decompressobj(wbits)
-        self._compressor = zlib.compressobj(level, zlib.DEFLATED, wbits)
+        self._decompressor = zlib.decompressobj(-zlib.MAX_WBITS)
         self._bootstrapped = None
         self._socket = socket
 
@@ -38,8 +36,13 @@ class DeflateSocket(object):
         return uncompressed
 
     def send(self, data):
-        if isinstance(data, memoryview):
-            data = data.tobytes()
-        chunk = self._compressor.compress(data)
-        self._socket.send(chunk + self._compressor.flush(zlib.Z_SYNC_FLUSH))
-        return len(data)
+        return self._socket.send(data)
+
+
+class DeflateEncoder(object):
+
+    def __init__(self, level):
+        self._compressor = zlib.compressobj(level, zlib.DEFLATED, -zlib.MAX_WBITS)
+
+    def encode(self, data):
+        return self._compressor.compress(data) + self._compressor.flush(zlib.Z_SYNC_FLUSH)
--- a/nsq/async.py
+++ b/nsq/async.py
@@ -15,9 +15,9 @@ except ImportError:
     ssl = None  # pyflakes.ignore
 
 try:
-    from .snappy_socket import SnappySocket
+    from .snappy_socket import SnappySocket, SnappyEncoder
 except ImportError:
-    SnappySocket = None  # pyflakes.ignore
+    SnappyEncoder = SnappySocket = None  # pyflakes.ignore
 
 try:
     import simplejson as json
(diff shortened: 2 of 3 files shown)


ploxiln:
lots of test failures due to:

```
nsq/async.py:18: in <module>
    from .snappy_socket import SnappySocket, SnappyEncoder
E     File "/home/travis/build/nsqio/pynsq/nsq/snappy_socket.py", line 46
E       return snappy.StreamCompressor().compress(data)
E                                                     ^
E   TabError: inconsistent use of tabs and spaces in indentation
```

nsq/deflate_socket.py
```diff
@@ -38,8 +38,14 @@ def _recv(self, size, method):
...
-        chunk = self._compressor.compress(data)
-        self._socket.send(chunk + self._compressor.flush(zlib.Z_SYNC_FLUSH))
-        return len(data)
+        return self._socket.send(data)
+
+
+class DeflateEncoder(object):
+
+    def __init__(self, level):
+        self._compressor = zlib.compressobj(level, zlib.DEFLATED, -zlib.MAX_WBITS)
```
mreiferson: we no longer need lines 10 and 12 in this file as a result of this moving here

mreiferson:
@SpaTmole thanks again, after addressing that last minor comment, would you mind squashing this down to 1 commit (no need to open a separate PR though)?

nsq/deflate_socket.py
```diff
@@ -38,8 +38,14 @@ def _recv(self, size, method):
         return uncompressed
 
     def send(self, data):
-        if isinstance(data, memoryview):
```
jehiah: I think we want to keep `memoryview` handling added in #191

jehiah: (or this logic can be moved into AsyncConn?)

SpaTmole: @jehiah Well, I guess it is not necessary anymore.
Since we compress the data in the layer above (before sending the data to the IOStream.write) it's no longer "our" responsibility, because compression will return just another bytestring which will be sent to `stream.write` and through tornado's functions calls untill the `socket.send` will eventually send it. 

https://github.com/tornadoweb/tornado/blob/master/tornado/iostream.py#L556 
Tells us that data might be both memoryview and bytes, so not a big deal if we send bytes.

The `send` method receives the result of `_command` call from https://github.com/nsqio/pynsq/blob/master/nsq/protocol.py#L82 
therefore it always will be bytes.

Please, tell me if I'm missing some cases which might bring the memoryview type into the compression level.

nsq/deflate_socket.py
SpaTmole: Done.

ploxiln:
you accidentally duplicated commits from the master branch when you rebased+squashed

nsq/deflate_socket.py
ploxiln: I think that the problem fixed in #191 is that the compressor did not handle a `memoryview`, the compressor required bytes, and depending on where the data came from (maybe another tornado read) it might be a memoryview. So we may need to convert memoryview to bytes before passing it to the compressor, either in `AsyncConn`:

```python
    def send(self, data):
        self.stream.write(self.encoder.encode(data))
```

or maybe just in the compressing encoders e.g. `SnappyEncoder`:

```python
    def encode(data):
        return snappy.StreamCompressor().compress(data)
```

ploxiln: ... I'll admit I'm not sure how `AsyncConn.send()` got a `memoryview` before if it always gets the result of `_command()`, but #191 says it did ...

ploxiln: ... oh I see, the previous abstraction was a Socket we give to a tornado IOStream, but now the compression is outside of that whole thing ... ok I see how this is all fine

ploxiln:
I took the liberty of re-rebasing this to remove the extraneous commits

mreiferson:
Cool, is this ready to go then? I'm confused at the conclusions around `memoryview`.

ploxiln:
... and this is where IOStream is making the memoryview: https://github.com/tornadoweb/tornado/blob/v4.5.3/tornado/iostream.py#L862

PR #210, opened 2018-07-05
fix high CPU load in DeflateSocket on nsqd disconnection

zexxonn:
- fixed handling messages from closed connection in DeflateSocket/SnappySocket recv()
- made DeflateSocket//SnappySocket recv() behaves as standart socket when connection is closed

If connection to nsqd is closed then reading from socket returns empty result and socket should be closed, otherwise it will lead to constant attempt to read from it, and to high cpu load.


mreiferson:
nice fix, thanks!

PR #214, opened 2018-07-09
remove remnants of python2.6 support

ploxiln:
see title

PR #215, opened 2018-07-13
add python 3 classifiers

aaronjheng:


ploxiln:
these classifiers were added just a few days ago in #214 and merged just now, sorry!

PR #217, opened 2018-07-20
fix auth_secret identify option on python3

ploxiln:
The root issue is that on python3 there are conflicting type assertions for auth_secret:

```python
class AsyncConn(...):
    def __init__(...):
         assert isinstance(auth_secret, string_types + (None.__class__,))
...
    def _on_response_continue(...):
        self.send(protocol.auth(self.auth_secret))

# in protocol.py ...

def auth(data):
    return _command(AUTH, data)

def _command(cmd, body, *params):
    assert isinstance(body, bytes_types), 'body must be a bytestring'
```

in #202, @ericb proposed just changing the initial type assertion to allow `bytes_types` instead of `string_types`. I'm fine with that, but I think it might be more "natural" to allow a string for `auth_secret` as well, so I've come up with my own variation using `to_bytes`.

(pushing just the test first to demonstrate failures, then the fix ...)

### DIFF
--- a/nsq/conn.py
+++ b/nsq/conn.py
@@ -7,8 +7,7 @@ import time
 import socket
 import logging
 
-from ._compat import string_types
-from ._compat import struct_l
+from ._compat import string_types, to_bytes, struct_l
 from .version import __version__
 
 try:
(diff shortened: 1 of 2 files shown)


ploxiln:
test isn't working at all on travis, dunno why, hang on a sec ...

ploxiln:
Ah - test doesn't pass on python2.7 where I would expect it to, because of `from __future import unicode_literals`, and the failure is impossible to understand because the exception is hidden in a different context and not captured, annoyingly enough. Anyway the fix should fix it.

ploxiln:
cc @jehiah

PR #223, opened 2018-10-12
Remove hardcoded tlsv1 version in ssl.wrap_socket

duczen:
Removed the hard coded TLSv1 version in `ssl.wrap_socket`. 

TLS options for a reader can be set with `tls_options`, however the ssl_version is already explicitly set which yields the following error when trying to set a different version:

` TypeError: wrap_socket() got multiple values for keyword argument 'ssl_version'`

This prevents me from using pynsq with `-tls-min-version=tls1.2` set on my nsqds. Removing the hard coded version allows setting this value as/if needed. Running without the hardcoded value, `nsq.Reader` with `tls_v1=True` and omitting `tls_options` correctly uses the min tls version set by nsqd.

### DIFF
--- a/nsq/conn.py
+++ b/nsq/conn.py
@@ -300,15 +300,17 @@ class AsyncConn(event.EventedMixin):
         #
         # first remove the event handler for the currently open socket
         # so that when we add the socket to the new SSLIOStream below,
-        # it can re-add the appropriate event handlers.
+        # it can re-add the appropriate event handlers. Default to TLSv1.2
+        # unless ssl_version is set otherwise.
         self.io_loop.remove_handler(self.socket.fileno())
 
         opts = {
             'cert_reqs': ssl.CERT_REQUIRED,
-            'ca_certs': default_ca_certs()
+            'ca_certs': default_ca_certs(),
+            'ssl_version': ssl.PROTOCOL_TLSv1_2
         }
         opts.update(options or {})
-        self.socket = ssl.wrap_socket(self.socket, ssl_version=ssl.PROTOCOL_TLSv1,
+        self.socket = ssl.wrap_socket(self.socket,
                                       do_handshake_on_connect=False, **opts)
 
         self.stream = tornado.iostream.SSLIOStream(self.socket, io_loop=self.io_loop)


ploxiln:
Hmm - yeah, `TLSv1` is not great, it does not allow TLSv1.1 or TLSv1.2. The default ("SSLv23" in python2.7 and "TLS" in python3.5) theoretically allows SSLv3 or any version of TLS. I suppose we wanted to explicitly disallow SSLv3. But today the only thing that makes sense besides the default would be TLSv1.2.

I'm fine with the default. Alternatively, we may want to just add `PROTOCOL_TLSv1_2` to `opts` just if `ssl_version` is not already present there. cc @mreiferson @jehiah

mreiferson:
> I'm fine with the default. Alternatively, we may want to just add PROTOCOL_TLSv1_2 to opts just if ssl_version is not already present there.

This sounds reasonable, care to make this change @duczen?

Thanks for the PR!

nsq/conn.py
```diff
@@ -300,15 +300,19 @@ def upgrade_to_tls(self, options=None):
...
         self.io_loop.remove_handler(self.socket.fileno())
 
         opts = {
             'cert_reqs': ssl.CERT_REQUIRED,
             'ca_certs': default_ca_certs()
         }
         opts.update(options or {})
-        self.socket = ssl.wrap_socket(self.socket, ssl_version=ssl.PROTOCOL_TLSv1,
+        if not opts.get('ssl_version'):
+            opts['ssl_version'] = ssl.PROTOCOL_TLSv1_2
```
ploxiln: there's a slightly more elegant way to do this - if you include `ssl_version` in the `opts` dict initially (just above, next to `ca_certs`), then `opts.update(options)` can override it

(sorry for not seeing this earlier)

duczen: Ah, yes. I like that much better. I'll have that up in a sec.

ploxiln:
Nice, thanks.

One last thing - can you squash your commits into one? (We prefer the commits to be manually squashed, and then we use the "Create a merge commit" option to record who merged.)

PR #225, opened 2018-10-28
tests: explicitly install flake8 3.6.0 on travis-ci

ploxiln:
A new version of flake8 was recently released, and new warnings
caused the test-suite to fail. This explicitly installs 3.6.0 so
after the latest warnings are fixed, this won't happen again
unexpectedly.

Also re-arrange travis_test.sh a bit.


ploxiln:
cc @jehiah @mreiferson

mreiferson:
thx

PR #227, opened 2019-01-24
docs: add pydoc strings for Writer methods

ploxiln:
Writer documentation notably does not currently include pub(),
mpub(), dpub(), connect_to_nsqd(), etc - https://pynsq.readthedocs.io/en/latest/writer.html

~~I added `:undoc-members:` to reader.rst for symmetry, but it doesn't actually have any public members without pydoc strings.~~

The alternative way to fix is to add pydoc strings for all Writer public methods (updated to do this).

cc @jehiah @mreiferson

### DIFF
--- a/nsq/writer.py
+++ b/nsq/writer.py
@@ -55,12 +55,12 @@ class Writer(Client):
 
             def get(self):
                 topic = 'log'
-                msg = 'Hello world'
-                msg_cn = 'Hello 世界'
+                msg = b'Hello world'
+                msg_cn = u'Hello 世界'.encode('utf-8')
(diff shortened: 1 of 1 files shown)


mreiferson:
> I guess the alternative way to fix is to add pydoc strings for all Writer public methods...

Yes, let's just do that instead?

ploxiln:
updated - I didn't document `writer.connect()` or `writer.connect_to_nsqd()` because they seem to be just for internal use by `_run()` and `__init__()`

mreiferson:
nice thx

PR #230, opened 2019-02-25
reader: in send_rdy(), don't enforce total_rdy <= max_in_flight

alpaker:
Naively enforcing that invariant in `_send_rdy()` prevents us from correctly updating each conn's `rdy` after reducing `max_in_flight`.  Additionally, there's no need for this check in `_send_rdy()`, as all callers have logic that's supposed to determine an appropriate rdy value.


andyxning:
Thanks @alpaker for check and refactor this code!

PR #185, opened 2017-06-07
PeriodicCallback: remove io_loop argument for tornado v5 compatibility

jehiah:
This removes the io_loop argument from `PeriodicCallback` for tornado v5 forwards compatibility.

WIP: This should ideally get tested against tornadov5 but i'm not sure an easy way to do that since it's not released yet. Also i'm not sure this is the complete set of changes needed yet.

```
Traceback (most recent call last):
  File "queuereader_events.py", line 32, in <module>
    only_include_hosts=settings.get('nsqd_silod_on_staging'),
  File "/.../local/lib/python2.7/site-packages/libbitly/NSQReader.py", line 77, in __init__
    super(Reader, self).__init__(*args, **kwargs)
  File "/.../local/lib/python2.7/site-packages/nsq/reader.py", line 159, in __init__
    super(Reader, self).__init__(**kwargs)
  File "/.../local/lib/python2.7/site-packages/nsq/client.py", line 19, in __init__
    io_loop=self.io_loop).start()
TypeError: __init__() got an unexpected keyword argument 'io_loop'
```

PR #232, opened 2019-03-18
tornado 5.x support

ploxiln:
addresses #185 and #203 

~~still adapting tests ... they passed in a mocked io_loop so it's a pretty big change if they can't do that~~ adapted

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,36 @@
 dist: xenial
 language: python
+
 matrix:
   include:
   - python: "2.7"
-    env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2
-  - python: "2.7"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2
+    env: TORNADO_VERSION=4.1   NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
   - python: "2.7"
     env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
-  - python: "2.7"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
   - python: "2.7"
     env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
+  - python: "2.7"
+    env: TORNADO_VERSION=5.1.1 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
 
   - python: "3.4"
-    env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2
+    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
   - python: "3.4"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
+    env: TORNADO_VERSION=5.1.1 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
 
   - python: "3.5"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2
+    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
   - python: "3.5"
-    env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
-  - python: "3.5"
-    env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
+    env: TORNADO_VERSION=5.1.1 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
 
   - python: "3.6"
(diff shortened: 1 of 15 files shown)


ploxiln:
note that I removed nsq-0.3.8 and tornado-4.4 from the test matrix, figured it was time

ploxiln:
`test_backoff.py` had all tests indented ... diff best viewed by ignoring whitespace changes

jehiah:
> note that I removed nsq-0.3.8 and tornado-4.4 from the test matrix, figured it was time

For what it's worth, my last tornado upgrade effort only got me as far as `4.1.0-patched` ;)

ploxiln:
hah, so does that mean we should try for compatibility down to tornado-4.0.z ...

I just found `tornado.iostream.IOStream.start_tls()` for upgrading a plain IOStream to an SSLIOStream, and it was added in 4.0 so it's theoretically possible.

jehiah:
I mean if 4.1+ doesn't cause any specific complications i'd appreciate it.

ploxiln:
I think this is feature-complete. Please take a look. If the direction looks good, I'll squash down to about 4 commits, and try to actually use it ...

nsq/conn.py
```diff
@@ -297,28 +286,21 @@ def send(self, data):
...
-            'ca_certs': default_ca_certs(),
             'ssl_version': ssl.PROTOCOL_TLSv1_2
         }
         opts.update(options or {})
-        self.socket = ssl.wrap_socket(self.socket,
-                                      do_handshake_on_connect=False, **opts)
 
-        self.stream = tornado.iostream.SSLIOStream(self.socket, io_loop=self.io_loop)
-        self.stream.set_close_callback(self._socket_close)
+        fut = self.stream.start_tls(False, ssl_options=opts, server_hostname=self.host)
```
ploxiln: I don't think we were checking if the TLS server certificate was valid for the hostname connected to. Now, if not `ssl.CERT_NONE`, this should check hostname ...

ploxiln:
mostly squashed

ploxiln:
I threw up a version bump to 0.8.4b1 because it makes testing with my systems a bit easier.
(Maybe 0.9.0b1 would be more appropriate.)

setup.py
```diff
@@ -16,7 +16,7 @@ def run_tests(self):
 
 
 # also update in nsq/version.py
-version = '0.8.3'
+version = '0.8.4b1'
```
jehiah: I agree this should probably be `0.9.0-beta1`

ploxiln: ok, will update ... I did the "b1" here and the "-beta1" in version.py because "b1" is more conventional for pypi, see for example https://github.com/tornadoweb/tornado/commit/7457d6f58387a801fa2c648daa341e7043734219

but I can also try "-beta1" here and see how that works for pip, probably does

jehiah: 👍 i don't have a strong preference on that specific detail

ploxiln: cool, I'll keep with the scheme (b1 here)
updated, will test a bit on my stuff over the next day or two

ploxiln:
I'm finally running this for my staging services ... it seems to be working just fine. But note that I don't really use compression or tls features.

.travis.yml
```diff
@@ -1,39 +1,36 @@
...
     env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
+  - python: "3.4"
+    env: TORNADO_VERSION=5.1.1 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
 
-  - python: "3.5"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2
-  - python: "3.5"
-    env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
   - python: "3.5"
     env: TORNADO_VERSION=4.4.3 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
```
mreiferson: should be tornado `4.5.3` with nsq `1.0.0-compat`

.travis.yml
```diff
@@ -1,39 +1,36 @@
...
+  - python: "3.5"
+    env: TORNADO_VERSION=5.1.1 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
```
mreiferson: should be nsq `1.1.0`

.travis.yml
```diff
@@ -1,39 +1,36 @@
...
 
   - python: "3.6"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2
+    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.1.0.linux-amd64.go1.10.3
```
mreiferson: should be nsq `1.0.0-compat`

.travis.yml
```diff
@@ -1,39 +1,36 @@
...
   - python: "3.6"
-    env: TORNADO_VERSION=4.5.3 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
+    env: TORNADO_VERSION=5.1.1 NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8
```
mreiferson: should be nsq `1.1.0`

nsq/deflate_socket.py
```diff
@@ -24,6 +24,17 @@ def recv(self, size):
     def read(self, size):
         return self._recv(size, self._socket.read)
 
+    def recv_into(self, buf, nbytes=0):
```
mreiferson: why is this method necessary?

ploxiln: it's the socket method with tornado-5.x uses instead of `recv()`, for potential efficiency improvement: https://github.com/tornadoweb/tornado/pull/2251/files#diff-46c2c62b2b1964831d9332ba934cb623R1185

.travis.yml
ploxiln: sure, but, fwiw, I was trying to mix it up so so that e.g. 5.1.1 was tested with nsqd-1.0.0 in some case and 4.5.3 was tested with nsqd-1.1.0 in some case (without expanding the test matrix too much)

ploxiln:
(I'll re-squash down to 3 or 4 commits, when ready)

mreiferson:
cool, error handling improvements look good 👍🏻

:hammer:

ploxiln:
mostly squashed

PR #234, opened 2019-05-07
我不知道如何理解

davidli006:
line 99  and line 134


ploxiln:
this is asking a question:

```python
class Writer():
    def __init__():
        ...
        self.conns = {} # How do we understand its relationship to line 134?`
```

and this is line 134:

```python
    def mpub():
        ...
        if isinstance(msg, bytes_types):  # line 134
```

and I don't understand this question at all.

We can only provide support in English here, sorry.

PR #238, opened 2019-08-25
enable ipv6 support

andyxning:
This PR adds support for ipv6 nsqd ip.


ploxiln:
I think that full ipv6 support would involve resolving hostnames manually and trying to connect to both ipv4 and ipv6 addresses that result. But this does fix handling of ipv6 literal addresses while not breaking anything else.

PR #241, opened 2019-11-08
modify message body type comment, string => bytes

shuoli84:
refer to https://stackoverflow.com/a/5901825/334999. 
2.x also has a type bytes which is just an alias to str, so "bytes" is a proper type for message.body attribute.


ploxiln:
thanks

PR #243, opened 2020-01-15
reduce duplicate rdy update requests

andyxning:
This PR will reduct duplicate rdy update requests.

### DIFF
--- a/tests/test_backoff.py
+++ b/tests/test_backoff.py
@@ -260,6 +260,7 @@ def test_backoff_many_conns(mock_ioloop_current):
     for i in range(num_conns):
         conn = get_conn(r)
         conn.expected_args = [b'SUB test test\n', b'RDY 1\n']
+        conn.last_exp_rdy = b'RDY 1\n'
         conn.fails = 0
         conns.append(conn)
 
@@ -271,7 +272,7 @@ def test_backoff_many_conns(mock_ioloop_current):
         msg = send_message(conn)
 
         if r.backoff_timer.get_interval() == 0:
-            conn.expected_args.append(b'RDY 1\n')
+            add_exp_rdy(conn, b'RDY 1\n')
 
         if fail or not conn.fails:
             msg.trigger(event.REQUEUE, message=msg)
@@ -279,7 +280,7 @@ def test_backoff_many_conns(mock_ioloop_current):
             conn.fails += 1
 
             for c in conns:
-                c.expected_args.append(b'RDY 0\n')
+                add_exp_rdy(c, b'RDY 0\n')
             conn.expected_args.append(b'REQ 1234 0\n')
         else:
             msg.trigger(event.FINISH, message=msg)
@@ -287,7 +288,7 @@ def test_backoff_many_conns(mock_ioloop_current):
             conn.fails -= 1
 
             for c in conns:
-                c.expected_args.append(b'RDY 0\n')
+                add_exp_rdy(c, b'RDY 0\n')
             conn.expected_args.append(b'FIN 1234\n')
 
         assert r.backoff_block is True
@@ -299,7 +300,7 @@ def test_backoff_many_conns(mock_ioloop_current):
             conn = timeout_args[1]()
             last_timeout_time = timeout_args[0]
         assert r.backoff_block is False
-        conn.expected_args.append(b'RDY 1\n')
+        add_exp_rdy(conn, b'RDY 1\n')
 
         fail = True
         if random.random() < 0.3 and total_fails > 1:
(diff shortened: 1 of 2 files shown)


andyxning:
@ploxiln @mreiferson

tests/test_backoff.py
```diff
@@ -345,9 +345,32 @@ def test_backoff_many_conns(mock_ioloop_current):
...
+                j = j + 1
+                continue
+            rdy_found = False
+            for index in range(len(filtered_expected_args) - 1, -1, -1):
+                arg = filtered_expected_args[index]
+                if arg.startswith(b'RDY'):
+                    rdy_found = True
+                    if arg != c.expected_args[i]:
+                        if not (len(
+                                filtered_expected_args) == 1 and c.expected_args[i] == b'RDY 0\n'):
```
ploxiln: This code to generate the expected output is ... too complicated. It is a too hard to read it and convince yourself that the test is basically right.

I can see how you got here. The assumption is that the library code is simple and correct, so the failing test needs to be fixed. But you went down the wrong path. I saw about 8 attempts pushed up over about 8 hours as you were working on this ... good evidence that this strategy is too complicated. There has to be a better way.

Maybe, instead of fixing-up `expected_args` at the end, the fixes should be while appending to `expected_args` throughout the test. But also, maybe the test setup is a bit "buggy" - this test does tricky mocking of the ioloop.

ploxiln:
I'm not convinced about the benefit of this change. How often does the overall "ready" algorithm generate redundant RDY messages, and is this the simplest way to improve it? Maybe, but I'm not convinced.

tests/test_backoff.py
andyxning: Yes. It took me about 10 hours to fix the test. I will take a look at this to make it more readable.

andyxning:
> I'm not convinced about the benefit of this change. How often does the overall "ready" algorithm generate redundant RDY messages, and is this the simplest way to improve it? Maybe, but I'm not convinced.

When:
* the max_in_flight setting to 4000
* the nsqd instances is about 1500

When consumer connects to every nsqd, it needs to rebalance the rdy around all the existing connections, at this time, a connection may send continuous duplicate rdy value.

andyxning:
@ploxiln You can just take a look at the go-nsq change. Its change is simpler. https://github.com/nsqio/go-nsq/pull/282

andyxning:
@ploxiln Code has been cleaned in the correct way. PTAL.

andyxning:
@ploxiln This PR's code has been cleaned. PTAL.~

ploxiln:
here's an idea for an even simpler test change:

```diff
--- a/tests/test_backoff.py
+++ b/tests/test_backoff.py
@@ -260,9 +260,15 @@ def test_backoff_many_conns(mock_ioloop_current):
     for i in range(num_conns):
         conn = get_conn(r)
         conn.expected_args = [b'SUB test test\n', b'RDY 1\n']
+        conn.last_exp_rdy = b'RDY 1\n'
         conn.fails = 0
         conns.append(conn)
 
+    def add_exp_rdy(conn, rdy):
+        if conn.last_exp_rdy != rdy:
+            conn.expected_args.append(rdy)
+            conn.last_exp_rdy = rdy
+
     fail = True
     total_fails = 0
     last_timeout_time = 0
@@ -271,7 +277,7 @@ def test_backoff_many_conns(mock_ioloop_current):
         msg = send_message(conn)
 
         if r.backoff_timer.get_interval() == 0:
-            conn.expected_args.append(b'RDY 1\n')
+            add_exp_rdy(conn, b'RDY 1\n')
 
         if fail or not conn.fails:
             msg.trigger(event.REQUEUE, message=msg)
@@ -279,7 +285,7 @@ def test_backoff_many_conns(mock_ioloop_current):
             conn.fails += 1
 
             for c in conns:
-                c.expected_args.append(b'RDY 0\n')
+                add_exp_rdy(c, b'RDY 0\n')
             conn.expected_args.append(b'REQ 1234 0\n')
         else:
             msg.trigger(event.FINISH, message=msg)
@@ -287,7 +293,7 @@ def test_backoff_many_conns(mock_ioloop_current):
             conn.fails -= 1
 
             for c in conns:
-                c.expected_args.append(b'RDY 0\n')
+                add_exp_rdy(c, b'RDY 0\n')
             conn.expected_args.append(b'FIN 1234\n')
...
```

andyxning:
@ploxiln code format is done. PTAL.

ploxiln:
cool. thanks.

andyxning:
@ploxiln Thanks for your review.

PR #245, opened 2020-01-19
add Writer.close() and Client.close(), improve Reader.close()

### LINKED ISSUE #196: reader: not able to stop a reader when no nsqlookupd address set
The reader's close method will call on_connection_closed, and if nsqlookupd address not set, it will try to reconnect to nsqd after 15 seconds, which means the reader can't be closed in this case?

```
        # in _on_connection_close()

        if not self.lookupd_http_addresses:
            # automatically reconnect to nsqd addresses when not using lookupd
            logger.info('[%s:%s] attempting to reconnect in 15s', conn.id, self.name)
            reconnect_callback = functools.partial(self.connect_to_nsqd,
                                                   host=conn.host, port=conn.port)
            self.io_loop.add_timeout(time.time() + 15, reconnect_callback)
```

ploxiln:
make Reader.close() prevent nsqd_tcp_addresses from reconnecting

Writer.close() also closes conns and stops from reconnecting

Client.close() stops the periodic _check_last_recv_timestamps()

fixes #196


mreiferson:
These changes seem fine, ready to land this @ploxiln?

ploxiln:
unfortunately I haven't tested this at all, I've just been using the tornado-6 branch

PR #247, opened 2020-04-24
tests:  switch from Travis-CI  to GitHub Actions for CI

ploxiln:
because Travis-CI no longer updates commit status reliably for the travis-ci.org integration (though it does seem to be working on this repo for now? I've seen it stop working on a bunch of my own, and also https://github.com/nsqio/go-nsq/pull/294 and https://github.com/nsqio/go-diskqueue/pull/15)

(and because GitHub is has lots of money for free CI servers, so they are a bit more available and reliable)


ploxiln:
hmm... I copied this from my own repos, where just adding this yml file in this dir seemed to kick it off ... but now it's not doing anything ...

ploxiln:
I think it's because the branch which introduces it must be in this repo `nsqio/pynsq`, rather than the separate `ploxiln/pynsq` ... I'll switch to working on this in #248

ploxiln:
huh it's working here now?

ploxiln:
GitHub Actions does not have an "include just these combos" feature like we used with Travis-CI, it only has an "exclude these combos" setting. But it gives lots of parallel instances so it's pretty fast even with a huge matrix. (The max is ~255.)

The effective test matrix is updated:
  * python versions 2.7, 3.5, 3.8 (added 3.8, removed 3.4 and between 3.5 and 3.8)
  * tornado versions 4.5.3, 5.1.1 (removed 4.1, 4.4)
  * nsq versions 1.0, 1.1, 1.2 (added 1.2)

You don't have to use a container with GitHub Actions, you can just ask for a particular version of python and get a workable environment straight on the VM. But I'm a bit worried about how fast github removes support for older python versions, e.g. python-3.4 was removed late last year, and can no longer be used. So I use public/official containers on DockerHub which I expect will continue to exist for years even if no longer updated. (I have tested with a python-3.4 container in one  of my own projects, but it's based on debian-9 stretch instead of debian-10 buster, and I figured it wasn't quite worth the mess here.)

If you want me to add older tornado or python-3.4 let me know :)

cc @jehiah

ploxiln:
... updated the strategy: include all python versions in test matrix, but do not include nsqd versions in test matrix, instead pick nsqd version based on python version

mreiferson:
Should we land this @ploxiln?

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
+name: CI
+
+on:
+  push:         {branches: [master]}
+  pull_request: {branches: [master]}
```
mreiferson, line 5: Out of curiosity, why this YAML syntax with explicit dictionary?

ploxiln: The only reason was to experiment with writing in a more concise format (one line with three words instead of 3 lines with one word). I have no problem switching it to the more typical format.

ploxiln:
I've been on a trip since friday, will be back a bit on sunday, but fully on monday. I'll make sure this is up-to-date and run it again before merge.

mreiferson:
SGTM

.github/workflows/main.yml
mreiferson: Yes, please 🙏

.github/workflows/main.yml
```diff
@@ -0,0 +1,60 @@
...
+    container: "${{matrix.imgtag}}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download NSQ
+        run: |
+          PY_VER="$(python -c 'import sys; print(sys.version_info[:2])')"
+          case "$PY_VER" in
+              '(3, 4)' | '(3, 7)' ) NSQ_DIST="nsq-1.0.0-compat.linux-amd64.go1.8" ;;
+              '(2, 7)' | '(3, 6)' ) NSQ_DIST="nsq-1.1.0.linux-amd64.go1.10.3" ;;
+              '(3, 5)' | '(3, 8)' ) NSQ_DIST="nsq-1.2.0.linux-amd64.go1.12.9" ;;
```
ploxiln, line 38: notice my weird scheme to reduce the text matrix by correlating version of python and version of nsqd

ploxiln:
updated

.github/workflows/main.yml
mreiferson: I did notice that, I resisted commenting 😁

mreiferson:
I guess we should follow up with `go-nsq` and `nsq`?

PR #263, opened 2022-04-18
pynsq and asyncio together

vortex14:
Many people use asyncio to work in python 3, added 2 new examples to help understand.


mreiferson:
Thanks for this, let's add the example source, but I don't think we need the update in the README. The CI linter is also complaining about trailing newlines...

vortex14:
> Thanks for this, let's add the example source, but I don't think we need the update in the README. The CI linter is also complaining about trailing newlines...

Already rollbacked README.md

mreiferson:
Great, would you mind squashing these commits down to one, please?

vortex14:
> Great, would you mind squashing these commits down to one, please?

ready !

